PR #13, opened 2017-02-07
Blocking util for testing and config refactor

adamstruck:


src/tes/logger/log.go
```diff
@@ -22,6 +23,26 @@ type Logger interface {
 	WithFields(...interface{}) Logger
 }
 
+// SetLogLevel sets the level of logging
+func SetLogLevel(l string) {
```
buchanae: Can we name it `SetLevel` please?

src/tes/scheduler/condor/sched.go
```diff
@@ -77,7 +73,7 @@ func (s *scheduler) startWorker(workerID string) {
 	submitTpl, _ := template.New("condor.submit").Parse(`
 		universe    = vanilla
 		executable  = {{.Executable}}
-		arguments   = -config worker.conf.yml
+		arguments   = -config {{.WorkDir}}/worker.conf.yml
```
buchanae: Is this a bug fix?

src/tes/config.go
```diff
@@ -26,44 +50,114 @@ type Openstack struct {
...
+	WorkDir      string
+	LogLevel     string
 }
 
 // DefaultConfig returns configuration with simple defaults.
 func DefaultConfig() Config {
 	workDir := "tes-work-dir"
 	return Config{
-		ServerConfig: pbr.ServerConfig{
+		ServerConfig: ServerConfig{
```
buchanae: Can we name it `Server` instead of `ServerConfig`? We know everything in these structs is `Config`

buchanae: Actually, since we've removed the protobuf messages, we can probably just ditch the `ServerConfig` struct and move all the fields to the root `Config` struct.

src/tes/config.go
```diff
@@ -26,44 +50,114 @@ type Openstack struct {
...
+	// Address of the scheduler, e.g. "1.2.3.4:9090"
+	ServerAddress string
+	// Directory to write job files to
+	WorkDir string
+	// How long (seconds) to wait before tearing down an inactive worker
+	// Default, -1, indicates to tear down the worker immediately after completing
+	// its job
+	Timeout int
+	// How often (milliseconds) the worker polls for cancellation requests
+	StatusPollRate int
```
buchanae: Wouldn't `time.Duration` be better than `int`?

src/tes/logger/log.go
```diff
@@ -22,6 +23,26 @@ type Logger interface {
...
+	case "info":
+		logrus.SetLevel(logrus.InfoLevel)
+	case "warn":
+		logrus.SetLevel(logrus.WarnLevel)
+	case "error":
+		logrus.SetLevel(logrus.ErrorLevel)
+	case "fatal":
+		logrus.SetLevel(logrus.FatalLevel)
+	case "panic":
+		logrus.SetLevel(logrus.PanicLevel)
```
buchanae: I purposefully left out warn/fatal/panic error levels from the logging API, so we can remove those levels here.

src/tes/logger/log.go
```diff
@@ -22,6 +23,26 @@ type Logger interface {
...
+	default:
+		logrus.SetLevel(logrus.DebugLevel)
```
buchanae: In linux commands you usually opt-in to debug logging, so maybe a better default is info

src/tes/scheduler/scheduler.go
```diff
@@ -28,7 +28,7 @@ type Scheduler interface {
 // Offers which aren't marked as rejected by the scheduler are accepted
 // and the job is assigned to a worker in the database.
 func StartScheduling(db *server.TaskBolt, sched Scheduler) {
-	ticker := time.NewTicker(time.Second * 5)
+	ticker := time.NewTicker(time.Millisecond * 10)
```
buchanae: Mistake here? I think this moved to config?

src/tes/server/task_boltdb.go
```diff
@@ -343,11 +344,10 @@ func (taskBolt *TaskBolt) GetServiceInfo(ctx context.Context, info *ga4gh_task_e
 	//     For example, you can't have multiple S3 endpoints
 	out := map[string]string{}
 	for _, i := range taskBolt.serverConfig.Storage {
-		if i.Local != nil {
+		if reflect.DeepEqual(i.Local, tes.LocalStorage{}) {
```
buchanae: This could do `len(i.Local.AllowedDirs) > 0` and avoid reflect and deep equal and the creation of a throwaway struct.

src/tes/server/task_boltdb.go
```diff
@@ -343,11 +344,10 @@ func (taskBolt *TaskBolt) GetServiceInfo(ctx context.Context, info *ga4gh_task_e
...
 			out["Local.AllowedDirs"] = strings.Join(i.Local.AllowedDirs, ",")
 		}
-
-		if i.S3 != nil {
+		if reflect.DeepEqual(i.S3, tes.S3Storage{}) {
```
buchanae: This could do `i.S3.Endpoint != ""`. 

Optionally it could also check the Key and Secret fields, which would provide some extra config validation. I don't know if key and secret are ever optional in S3 though.

src/tes/storage/storage.go
```diff
@@ -89,26 +90,28 @@ func (storage Storage) WithLocal(allow []string) (*Storage, error) {
...
+	if reflect.DeepEqual(conf.Local, tes.LocalStorage{}) {
 		out, err = storage.WithLocal(conf.Local.AllowedDirs)
+		if err != nil {
+			return nil, err
+		}
 	}
 
-	if conf.S3 != nil {
-		// TODO need config validation to ensure these values actually exist
+	if reflect.DeepEqual(conf.S3, tes.S3Storage{}) {
```
buchanae: Similar to other comments, checking `conf.S3.Endpoint != ""` would be fine (better IMO)

src/tes/worker/engine.go
```diff
@@ -25,14 +26,14 @@ type Engine interface {
 
 // engine is the internal implementation of a docker job engine.
 type engine struct {
-	conf Config
+	conf tes.Worker
```
buchanae: I'm being picky here, but `tes.Worker` doesn't read like a config type. So I'm contradicting my earlier comment about ServerConfig.

Possibly the most idiomatic Go approach would be to move the config stuff to a `config` package, then all config structs are referenced e.g. `config.Worker`

src/tes/worker/engine.go
```diff
@@ -77,7 +78,9 @@ func (eng *engine) RunJob(parentCtx context.Context, jobR *pbr.JobResponse) erro
 	}()
 
 	// Ticker for State polling
-	tickChan := time.NewTicker(time.Second * 5).C
+	pollRate := time.Duration(eng.conf.StatusPollRate)
+	tickChan := time.NewTicker(pollRate * time.Millisecond).C
```
buchanae: Making the config struct have type of `time.Duration` would mean we don't need these lines to convert it from an int.

adamstruck:
closing since #14 contains these changes.

PR #15, opened 2017-02-13
Smarter schedulers

buchanae:
The schedulers are currently very limited; they don't have any concept of worker resources or limits, and mostly they just submit jobs to the first worker found without bound, or have some arbitrary bound.

The goal here is to make the schedulers smarter. In the process, the communication between the worker, server, database, and scheduler will change, to allow the schedulers to have the information they need about available workers. Also, hopefully the worker will be simplified in the process.

Putting this work up now to get eyes on it early, as it will be a big change.

Side note, I'm also adding "DO NOT MERGE" and "Ready for merge" labels to the repo. Please watch out for these before merging PRs.


buchanae:
I added some code for filtering and scoring workers in the last commit. Still very proof-of-concept, but the core idea is there. Would love to know what you think! Scores + weights seems could be fairly powerful.

Also removed the Offer struct and Accept/Reject/Wait mechanism I had previously dreamed up. Things may be simpler without it and it's one less concept to explain/understand. Instead of having these Offer state functions, a scheduler could instead just poll the database/server for jobs assigned to certain worker IDs, similar to what a worker would do. Schedulers probably need to do this polling anyway, to get fresh worker state/info.

Removed the multisched/coordinator for now, mainly so my code compiles. I'll put them back when things settle down.

buchanae:
The latest commit adds some simple "last ping time" checking for worker health and adds a state fields to the worker message in the DB (Dead/Alive). I don't actually delete workers from the database yet though, need to think about when it's ok to do that.

A manual test worked, showing the scheduler was updating the available worker resources, matching by CPU and RAM limits, and ignoring dead workers.

buchanae:
The latest commits are getting pretty close to done. I've added some mocking and testing in Go tests which have helped a lot with the GCE stuff.

There are also new helper scripts for deploying to GCE. That's in the ./gce dir.

This doesn't yet have a way to shutdown idle workers. I removed the idle timeout that previously existing in the worker and haven't added it back yet. I'd prefer to move this to the scheduler code, because from there a scheduler could also clean up cloud resources (e.g. delete GCE instances). I think I'd like to revisit that in a future PR, so that this one can be closer to done.

`make lint` is currently failing the build because it's complaining that I'm doing a "dot" import (import all names from a package). In the case where I'm using it, it's super handy, so I'd argue we find a way to disable that check. Otherwise, the tests are passing.

buchanae:
Some more things left to do (not necessarily blocking this PR):
- handle dead workers better. If they have jobs assigned, maybe re-queue? When to delete dead workers from the database?
- handle jobs that are never started. possibly because the worker is dead.
- graceful worker shutdown
- add workers to dashboard
- google storage directory download
- validate the config more, especially the google scheduler config. 
- generic cluster scheduler for case where you want to manually start workers
- there are lots of error conditions to test, which are much easier now with mocks.
- need a way to stop worker goroutines.

buchanae:
Another TODO:
a worker in the "Initializing" state that fails to start never gets marked as dead

buchanae:
I've added the worker timeout back and put the google storage directory download on a separate branch.

Ready to get this PR closed up, just waiting on review.

src/tes/worker/runner.go
```diff
@@ -11,225 +10,211 @@ import (
...
-			return err
-		}
-
-		stepRunners = append(stepRunners, s)
-	}
+	r.step("prepareDir", r.prepareDir)
+	r.step("prepareMapper", r.prepareMapper)
+	r.step("prepareStorage", r.prepareStorage)
+	// TODO prepareIP can fail when there is no network connection,
+	//      but should just return no IP. Fix and test.
```
adamstruck: Was this ever resolved?

buchanae: Not yet. Filed https://github.com/ohsu-comp-bio/funnel/issues/24

src/tes/storage/gs_test.go
```diff
@@ -0,0 +1,72 @@
...
+	var err error
+	var gs *GSBackend
+	gs, err = NewGSBackend(conf)
+	if err != nil {
+		t.Errorf("Can't get authed backend: %s", err)
+	}
+	return gs
+}
+
+func TestAnonymousGet(t *testing.T) {
```
buchanae: Fixed

.gitmodules
```diff
@@ -49,3 +49,18 @@
...
+	url = https://code.googlesource.com/gocloud
+[submodule "src/vendor/github.com/googleapis/gax-go"]
+	path = src/vendor/github.com/googleapis/gax-go
+	url = https://github.com/googleapis/gax-go
+[submodule "src/vendor/google.golang.org/api"]
+	path = src/vendor/google.golang.org/api
+	url = https://code.googlesource.com/google-api-go-client
+[submodule "src/vendor/github.com/shirou/gopsutil"]
+	path = src/vendor/github.com/shirou/gopsutil
+	url = https://github.com/shirou/gopsutil
```
buchanae: Done

share/index.html
```diff
@@ -28,7 +28,7 @@
```
buchanae: Fixed. I did a semi-hacky quick fix. We'll have to revisit the dashboard when we have more time.

Makefile
```diff
@@ -44,7 +46,8 @@ tidy:
 	@find ./* -type f | grep -E '.*\.py$$' | grep -v "/venv/" | grep -v "/share/node" | xargs autopep8 --in-place --aggressive --aggressive
 
 lint:
-	flake8 --exclude ./venv .
+	pip install -q flake8
+	flake8 --exclude ./venv,./share .
 	go get github.com/alecthomas/gometalinter
 	./buildtools/bin/gometalinter --install > /dev/null
 	./buildtools/bin/gometalinter --disable-all --enable=vet --enable=golint --enable=gofmt --vendor -s ga4gh -s proto ./src/...
```
buchanae: Done

Makefile
```diff
@@ -13,13 +14,14 @@ proto_build:
...
 		task_execution.proto
 	@cd proto && protoc \
 		$(PROTO_INC) \
 		-I ../task-execution-schemas/proto/ \
-		--go_out=Mtask_execution.proto=tes/ga4gh,plugins=grpc:../src/tes/server/proto \
+		--go_out=$(GRPC_HTTP_MOD),Mtask_execution.proto=tes/ga4gh,plugins=grpc:../src/tes/server/proto \
+		--grpc-gateway_out=logtostderr=true:../src/tes/server/proto/ \
 		task_worker.proto
 
 depends:
```
buchanae: Done with add_deps (vendored)

PR #16, opened 2017-02-15
Local storage refactor

adamstruck:
Quick fix for file permissions issue in containers with non-root users.


adamstruck:
@buchanae I think we should merge this for now and make the changes we talked about later on

PR #21, opened 2017-02-21
Webapp updates

adamstruck:
* use npm for dependencies
* sort, filter, pagination, cancel button on job list table


buchanae:
👍

buchanae:
Could you check that all dependencies are in package.json? I think maybe it's missing node-sass.

Also, you might want to refer directly to the binary in the Makefile, e.g. `./node_modules/.bin/node-sass`. Currently I think it's possibly referring to a global install.

PR #22, opened 2017-02-27
Hard Links

adamstruck:
* Hard link inputs when possible
    * input in read-only volume
    * local file system
* Handle symlinks more transparently

Depends on #15.


adamstruck:
Will resubmit with additional features related to use case work

PR #23, opened 2017-03-13
Gs dir download

buchanae:
Add get directory handling to google storage backend.

Wait for schedulers branch before merging.


buchanae:
I ended up needing this to fix tests in the smarter-sched branch.

PR #32, opened 2017-03-15
Readonly file linking

buchanae:
Credit goes to @adamstruck for this one. I just made some minor changes to code he wrote previously in another branch.

@adamstruck I got rid of the "ReadOnlyFile" class in favor of a readonly bool argument. It seemed simpler? Not sure.

Blocked on merging #15


adamstruck:
I had some python tests written for this that appear to fail when I run them on this branch.

PR #46, opened 2017-03-17
Set UID

adamstruck:
Is this the approach we want to take going forward? 

This PR will serve as a forum to address this question.


adamstruck:
This can break tools that use environmental variables that are affected by changing the user such as HOME. 

New approach we are going to sets is to set the GID on each output volume so that all output files created in the container have a GID that will allow the funnel worker to access them.

PR #54, opened 2017-03-24
Gce deploy

buchanae:
- Adds bash scripts to help deploy Funnel server/workers on GCE.
- Changes the way the GCE scheduler looks up instance templates so that it looks for a "funnel" tag, instead of expecting all templates to be set in the config. This makes it much easier to configure worker templates.


buchanae:
Ignore the bundling commits, those are from #52 (in this branch by accident)

Issue #50, opened 2017-03-23
Resolve symlinked outputs


buchanae:
There's a common case where tools will symlink files into an output directory, which Funnel then tries to upload. The upload fails since the symlinked path is relative to the container's filesystem and Funnel runs on the host filesystem.

We can improve this by doing some simple path rewriting to map the container path to a volume mounted on the host filesystem.

PR #57, opened 2017-03-27
Resolve sym

buchanae:
#50


buchanae:
Just realized this probably doesn't fix symlinks inside directory outputs, so adding "do not merge"

buchanae:
Latest commit should fix symlinks in directory outputs. I removed the Go test in favor of python tests for now, because it's currently too tricky to test the job runner in Go.

PR #61, opened 2017-04-03
funnel run command

buchanae:
This code is still a mess, so I'm not ready to merge it. Just want to get the code shared.


buchanae:
While trying to iron out some of the tricky details of this, I started thinking of a redesigned approach that is more similar to dsub. Here are some rough examples:

```bash
  funnel run 'md5sum' --stdin ~/input.txt --stdout gs://bkt/output.txt

  funnel run --script runscript.sh --in input ~/input.txt --out output gs://bkt/results.txt

  funnel run <<TPL
    'acme-aligner -f $factor -x $other -p1 $pair1 -p2 $pair2 -o $alignments'
    --in pair1 ~/pair1.fastq                        
    --in pair2 ~/pair2.fastq                        
    --contents other ~/chroms.txt                    
    --name pair1 'Pair 1 inputs'                    
    --desc pair1 'Pair 1 description'               
    --type pair1 directory                          
    --out alignments gs://bkt/bowtie/alignments.bam 
    --factor 5                                      
    --image my-bowtie2                              
    --name 'Bowtie2 test'                           
    --volume /tmp                                   
    --volume /opt                                   
    --tag 'foo: bar'
    --cpus 10
    --ram 50GB
    --disk 1TB
    --description 'Testings an example of using `funnel run` for a bowtie2 command'
TPL
```

This format allows for all the little details to be available (name, contents, resources) in a more consistent and obvious way.

There's also a rough sketch of how you might use this to stitch together a workflow:
https://github.com/buchanae/funnel/blob/tes-client/src/funnel/cmd/run/run.go#L273

PR #71, opened 2017-04-13
Move server and scheduler start code.

buchanae:
- Shuffles some server and scheduler code around to clean things up.
- Tries to simplify the cmd/server.go code needed to start a funnel server. Partially succeeds.
- Renames Scheduler to Backend, then creates Scheduler to represent high-level Scheduler server/controller
- Server/Scheduler create code is more easily reused by tests which create a mock server.
- Better use of context by scheduler
- Adds "Cache-Control: no-store" HTTP response header to prevent responses from being cached. This is configurable with config.DisableHTTPCaching (default true)
- Adds "make go-test" command for running the go tests only. The need for this command is another reason I'd like to move the code out of src/funnel.


buchanae:
This also takes one small step towards being able to create a pluggable database backend by relying on a separate Database interface for both Scheduler and Server, and requiring that database be passed in to NewServer and NewScheduler.

buchanae:
This also prunes the gorilla dependency, since all path mapping is now handled by core Go http library.

buchanae:
The latest commits:
- update the travis build to go 1.8
- Rename Factory to Plugin and add plugins directly to Scheduler instance via Scheduler.AddBackend instead of having a global plugin registry
- Clean up mock server tests. There were two mock servers, not there is one and the helper functions simplified some of the test code.

adamstruck:
:+1:

PR #74, opened 2017-04-14
Bundle examples

buchanae:
```
RJHB552 funnel (bundle-examples)
./bin/funnel examples task_message
{
    "name" : "Test Echo, Sleep, Echo",
    "projectId" : "Funnel",
    "description" : "Simple Sleep Command",
    "resources" : {
	"minimumCpuCores": 2,
        "volumes": [
	    {
		"name": "test_disk",
		"sizeGb": 5,
		"mountPoint": "/tmp"
            }
	]
    },
    "docker" : [
	{
	    "imageName" : "ubuntu",
	    "cmd" : [
		"bash",
		"-c",
		"echo foo && sleep 10 && echo bar"
	    ],
	    "stdout" : "stdout",
	    "stderr" : "stderr"
	}
    ]
}
```


buchanae:
Depends on #73

python/examples/md5-s3.py
```diff
@@ -0,0 +1,61 @@
+import argparse
```
adamstruck: This file needs to be updated or removed.

adamstruck:
Merging. #91 is related and will be addressed in a future PR.

PR #86, opened 2017-04-18
Gce auto start

buchanae:
Simplies the GCE bash scripts significantly and adds a `funnel gce start` command which allows funnel to read configuration from GCE instance metadata and automatically decide whether to start a server or worker based on configuration.

A couple other changes got mixed in here:
1. Move server/worker command files so that the code can be easily shared with the gce command
2. Fix the cache-control header for the GetJob endpoint
3. Regenerate and rename mock files
4. Refactor some config file loading code to provide access to parsing YAML from a string instead of a file.


buchanae:
There's at least one issue which needs a fix/test.

PR #95, opened 2017-04-26
Prototype terminal dashboard; includes updates to task cli

adamstruck:
Prototype.

Still needs:
* Better error handling
* Task view layout
* Pass golint

### DIFF
--- a/logger/log.go
+++ b/logger/log.go
@@ -41,6 +41,15 @@ func SetLevel(l string) {
 	}
 }
 
+// DisableTimestamp prevents timestamps from being displayed in the logs
+func DisableTimestamp(b bool) {
+	formatter := &textFormatter{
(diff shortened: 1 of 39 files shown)


adamstruck:
#96 tracks the additional improvements that need to be done in the future.

logger/log.go
```diff
@@ -8,7 +8,8 @@ import (
 )
 
 var formatter = &textFormatter{
-	DisableTimestamp: true,
+	DisableTimestamp: false,
+	FullTimestamp:    true,
```
buchanae: I don't want timestamps in my logs. Can we have a config/flag for this please? I don't care if the default is on or off, as long as I can easily set it.

adamstruck: It is now configurable via the `TimestampLogs` field in the config.

PR #98, opened 2017-04-27
Funnel run

adamstruck:
Picking up from #61


adamstruck:
This now works:

```
funnel run 'md5sum $FILE > $OUT' --in FILE=file:///tmp/file.txt --image ubuntu --out OUT=file:///tmp/test_out
```

buchanae:
- default server address should be "http://localhost:8000"
- remove "stringSlice" from help
- remove "vars-file"
- make image an arg
- help should include example of input, output, env var format e.g. "key=val"
- the flags shouldn't be sorted by name if possible, but rather grouped by concept (most common in/out/env, metadata, resources, etc)

buchanae:
- "--print" shows debug logs which threw me off. Also, pretty printing the JSON would be a nice default to have

buchanae:
- automatically prefix "file://"

buchanae:
- '--stdin' and '--stdout' flags would be useful
- directory input

adamstruck:
Remaining TODO's tracked in #105

Issue #105, opened 2017-04-28
Enhance funnel run


adamstruck:
- [x] stdin, stdout and stderr flags
- [x] variables file
- [x] command file
- [x] pretty print json with `--print` flag
- [x] sort flags logically, not alphabetically
- [x] task dependencies via `--wait`

buchanae:
- [ ] parse flags from bash heredoc e.g.

```
funnel run 'md5 $FILE' <<TPL
  -c ubuntu
  --in FILE=~/input.txt
TPL
```

buchanae:
These are in the help example
```
    --in pair1=file://~/pair1.fastq
    --in pair2=file://~/pair2.fastq
```

buchanae:
Other ideas:
- [ ] `--contents` for TaskParameter.contents
- [ ] `--script` for loading the command string from a file

buchanae:
- [ ] thorough examples on the website and in `funnel help`

jbingham:
What does --stdin do? 

stdout and stderr are the bucket paths to where the Docker command's stdout and stderr should be written, right?

buchanae:
> What does --stdin do?

It gets added to the input files, downloaded, and connected to the executor process's stdin

> stdout and stderr are the bucket paths to where the Docker command's stdout and stderr should be written, right?

Correct.

jbingham:
Re stdin: do you have a real-world example where being able to do that is that useful?

buchanae:
> Re stdin: do you have a real-world example where being able to do that is that useful?

No, it's supported mainly because it's part of TES Executors.

buchanae:
#110 takes care of most of this. Still a little bit of cleanup work to do there.

PR #110, opened 2017-05-10
Funnel run enhancements

buchanae:
- pretty print (indent) task json output
- ignore web-dashboard/web.go during `make tidy`
- wrote a better help/usage blob
- New flags:
```
      --wait-for    Wait for the given task IDs before running the task.
  -x, --extra       Parse funnel run arguments from the given string.
  -X, --extra-file  Parse funnel run arguments from the given file.
      --scatter     Scatter multiple tasks, one per row of the given file.
  -C, --contents    Include input file contents a file e.g. varname=/path/to/in.txt
```

- `logger.PrintSimpleError` basically prints "ERROR:" in red
- default container image of "alpine"

I haven't tested this code much, other than to run a few basic tests and look at the task JSON printed out. `--contents` isn't implemented in the server/worker yet.

Issue #105


cmd/run/cmd_test.go
```diff
@@ -0,0 +1,182 @@
...
+						"e2": "e2v",
+						"f1": "/opt/funnel/inputs" + cwd + "/testdata/f1.txt",
+						"f2": "/opt/funnel/inputs" + cwd + "/testdata/f2.txt",
+						"f4": "/opt/funnel/inputs" + cwd + "/testdata/f4",
+						"f3": "/opt/funnel/outputs" + cwd + "/testdata/f3",
+						"o9": "/opt/funnel/outputs" + cwd + "/testdata/o9",
+					},
+				},
+			},
+			Volumes: []string{"volone", "voltwo"},
```
adamstruck: these volumes need to be absolute paths

buchanae:
The latest commit adds a dead simple validation on volumes, checking that they being with a slash. I also updated the tests to pass absolute paths, but I'm not testing any funnel run validation/error handling in this PR since that's a lot more work.

buchanae:
Should all be contained in #122

PR #119, opened 2017-05-13
Fix webdash and add test

buchanae:
The web dashboard was broken on master from a recent PR. This fixes that and adds a test to make sure the webdash bundle is being served.


adamstruck:
you forgot to run `make tidy`

PR #121, opened 2017-05-17
Local storage fix

adamstruck:
Fix for case when we might encounter `Put` being called multiple times on the same output file.  

The issue was that first time `Put` was called it hard links the file. The second time its called hard linking fails since the dest already exists. This results in a fallback to `copyFile`  where `os.Create` is called on dest thus resulting in the truncation of the source file (since source and dest refer to the same inode).


buchanae:
My concern here is that these two Put() calls could be from two different tasks, right? What's the expected behavior of two tasks writing to the same file, sequentially? I would expect the second to overwrite the first, although I could also see a case for raising an error, but if I understand correctly, this code does neither of those options.

adamstruck:
os.SameFile is checking that the two inode references are the same.   If they are not then the file will be overwritten.

adamstruck:
test was updated to make things more clear.

Issue #18, opened 2017-02-16
Test annoyances


buchanae:
During testing, I frequently run into the same issues and because I'm running the tests dozens of times an hour, these get pretty annoying.

- [x] logs for all tests are all merged together with no clear separator.
- [x] the 'tes-wait' container doesn't stop and keeps the 5000 port locked, which causes failures
- [x] referring to a single test isn't simple: `nosetests tests/test_client.py:TestTaskREST.test_cancel`

buchanae:
I tried nose-capturestderr to break up the server logs, but it doesn't seem to capture a subprocesses stderr.

buchanae:
Another possible improvement: run all the tests in parallel against a single server. This is more realistic and would be faster and have less issues with starting/stopping a server per-test.

buchanae:
Fixes in #122 

...except for breaking up the log files per-test. I haven't figured out a solution for this, although it's much more manageable now that the default log level is info and it's easy to run a specific test with `go test ./path -run Regex`. Still, sometimes the FAIL lines get lost in the logs.

buchanae:
https://github.com/ohsu-comp-bio/funnel/pull/123/commits/4ccd94ae854d56ed750c1b71f889366123a89f09

Adds a utility which runs `go test -v` and transforms the output to add some spacing and color. It also prints a test suite summary at the end, so you don't have to hunt through the logs. It's still a bit rough around the edges, and only seems to work with `-v` (verbose) given to `go test`, but it's a good enough first pass for this issue.

buchanae:
In summary:
- tes-wait is gone, replaced (at least for now) by other forms of waiting that don't require another container
- the `go test` logs are parsed, colorized, etc. to make them more readable
- I tried running tests in parallel, but currently things break. We should look into why.
- The python tests are gone, so referring to a single test is easy, e.g. `go test ./tests/e2e/ -run Hardlink`

Issue #38, opened 2017-03-17
Add command-line interface


jbingham:
One thing we've found super useful at Google/Verily is to have a qsub-style command-line interface to the Pipelines API. We've got a version called dsub now at:

https://github.com/googlegenomics/task-submission-tools

We'd like to enhance dsub so that it supports TES as a backend. Then it could support local execution, job schedulers on HPC clusters, and multiple cloud providers.

Some possibilities:
- we could add Funnel as a backend to dsub. 
- we could fork dsub and add it as a CLI to Funnel
- we could contribute dsub to the ga4gh repo once it supports pluggable backends

buchanae:
A TES/funnel backend for dsub sounds great.

Ideally, I'd like to have a single funnel binary that includes something like dsub. That would go towards the goal of super easy download and usage.

A crazy idea (and probably separate issue): it would be amazing if somehow I could run `funnel run gcr.io/my-container samtools index gs://path/to/input.bam gs://path/to/output.bai`. I don't know how you'd determine which paths are outputs, but it might be possible.

jbingham:
The current dsub command-line would be:

dsub --project bill-to-project --logging gs://path/to/logs --image gcr.io/my-container --command "samtools index ${BAM} ${BAI}" --input BAM=gs://path/to/input.bam --output BAI=gs://path/to/output.bai

It's not too far off. With a few tweaks, the billing project could be scooped up from the user's environment. The logging path could default to the output path gs://path/to/.

buchanae:
I've been experimenting with a `funnel run` command here:
https://github.com/buchanae/funnel/tree/tes-client

The current syntax is something like:
`funnel run ubuntu 'md5sum {in .SRC} > {out .DST}' -- -SRC ~/path/to/input.txt -DST gs://bkt/out.txt`

This would send the task to server and print a job ID to stdout. There's a `--wait` flag for blocking until the job is done. There's a `--print` flag for printing the task JSON without running it. There are some flags for setting the resource requests (cpu, ram, disk, etc).

Being able to map/scatter over lines in a TSV/JSON/YAML file would be useful. That would print out multiple job IDs, one per line.

A `funnel wait` command might also be useful for waiting on a group of job IDs, which might give you more flexibility in throwing together ad hoc pipelines.

The code is still pretty rough and experimental and full of holes, but with a little work I think it can be pretty useful.

[...]

PR #122, opened 2017-05-18
Move python tests to Go

buchanae:
Based on the funnel run PR, so adding do not merge


buchanae:
The travis builds are 1-2 minutes faster for this branch, which is not as great as I was hoping, but still a nice improvement. Looks like the majority of our build time is in cloning submodules (40s), linting (25 s), building (20s), and testing (20-30s).

buchanae:
Took me awhile to figure out that the hardlink test in python was testing hardlinking of inputs, not outputs (which is what my converted Go test was for). 

Hardlinking outputs works on my mac, but not in travis nor exastack, because the file is owned by root (the container user). I tried to implement setgid, but it didn't help. In the end, I gave up.

buchanae:
I think we can close #18 #38 and #105 with this

adamstruck:
Looks good to me. 👍

PR #124, opened 2017-05-18
Feature/contents

adamstruck:


buchanae:
Code looks ok, but how does this pass input validation in the task runner? Needs a test before merging.

adamstruck:
`validateInputs` uses the inputs from the mapper.  When a file is created from the passed 'contents' that input is not added to mapper.Inputs. 

I'll write some tests to demonstrate the behavior.

buchanae:
Makes sense. Thanks.

adamstruck:
Added a test for the file mapper as a whole.

buchanae:
Could you add a ioutil.ReadFile to the test that checks the file that was created by task input contents?

adamstruck:
Done

PR #127, opened 2017-05-19
Remove protobuf types from config (#120)

### LINKED ISSUE #120: Protobuf types in config.Config causes inconsistent YAML keys
Config.Worker.Resources isn't being loaded from config correctly because it's a protobuf type (pbf.Resources). We

buchanae:
Fixes #120


adamstruck:
Please update examples/config.yml to match.

Issue #89, opened 2017-04-24
Possible to infer relative path for volume


buchanae:
Found while submitting this task:
```
{
    "name": "Test Google Storage",
    "project": "Funnel",
    "description": "Simple Echo Command",
    "resources": {},
    "executors": [
      {
        "image_name": "ubuntu",
        "cmd": ["md5sum", "/tmp/test_file"],
        "stdout": "stdout",
        "stderr": "stderr"
      }
    ],
    "inputs": [
      {
        "name": "infile",
        "description": "File to be MD5ed",
        "url": "gs://smc-rna-funnel/md5-input.txt",
        "type": "FILE",
        "path": "/tmp/test_file"
      }
    ],
    "outputs": [
      {
        "name": "outfile",
        "description": "MD5 output",
        "url": "gs://smc-rna-funnel/md5-output.txt",
        "type": "FILE",
        "path": "stdout"
      }
    ]
}
```

And the output. The key is in the stderr logs:
```
{
  "id": "b3v7vtfntf6g2u90l810",
  "state": "ERROR",
  "name": "Test Google Storage",
  "project": "Funnel",
  "description": "Simple Echo Command",
  "inputs": [
    {
      "name": "infile",
      "description": "File to be MD5ed",
      "url": "gs://smc-rna-funnel/md5-input.txt",
      "path": "/tmp/test_file"
    }
  ],
  "outputs": [
    {
      "name": "outfile",
      "description": "MD5 output",
      "url": "gs://smc-rna-funnel/md5-output.txt",
      "path": "stdout"
    }
  ],
  "resources": {},
  "executors": [
    {
      "image_name": "ubuntu",
      "cmd": [
        "md5sum",
        "/tmp/test_file",
        ">",
        "/tmp/md5_out"
      ],
      "stdout": "stdout",
      "stderr": "stderr"
    }
  ],
  "logs": [
    {
      "logs": [
        {
          "start_time": "2017-04-24T22:45:13Z",
          "end_time": "2017-04-24T22:45:13Z",
          "stderr": "invalid value \"/opt/funnel/funnel-work-dir/b3v7vtfntf6g2u90l810:.:rw\" for flag -v: .:rw is not an absolute path\nSee 'docker run --help'.\n",
          "exit_code": 2,
          "host_ip": "10.138.0.66"
        }
      ]
    }
  ]
}
```

buchanae:
```
[...]

PR #128, opened 2017-05-19
Task validation

### LINKED ISSUE #56: Task Message Validation
There are several fields in the [TES Schema](https://github.com/ga4gh/task-execution-schemas) that are documented as being required, but not treated as such by Funnel. 

Desired behavior: 
* message validation occurs after POST to /v1/jobs
    * 2xx response returned upon successful validation otherwise returns 4xx.

buchanae:
Closes #56 

Still needs a few more tests for individual validations.


buchanae:
Should also close #89 probably

PR #135, opened 2017-05-24
Update vendor packages

adamstruck:
Updates vendor package versions.


adamstruck:
Updates the grpc gateway which allows us to make calls like `curl -X GET http://localhost:8000/v1/tasks?view=BASIC`

PR #136, opened 2017-05-25
waitForRunning in TestMarkCompleteBug

buchanae:


adamstruck:
should `waitForRunning` be added before all instances of `waitForExec`?

buchanae:
I don't think it's necessary. This fixes the immediate problem, and other tests using waitForExec don't care about the task state. If it continues to be a problem, we could add waitForRunning inside waitForExec

adamstruck:
OK

👍

PR #141, opened 2017-05-31
set default AllowedDirs in the default config

adamstruck:


buchanae:
The website and example bundle need to be regenerated.

I think this should be squashed into a single commit.

A simple test in the storage package that checks that WithConfig results in the correct backends would be good.

adamstruck:
The website is unaffected since it serves the config example file. 

Otherwise, I updated the example bundle and added a test.

PR #142, opened 2017-06-01
cloud-init + google COS image

buchanae:
This simplifies deployment to Google by using their container-optimized OS (COS) and dockerized funnel. This removes the steps needed to build and deploy a Funnel VM image.

I'm leaving the existing VM-image deployment as is for now, in case it's actually useful to someone, some day. I'll revisit down the road to clean it up if it's not used.


adamstruck:
Could you add a README to the deployments section that describes the two sets of GCE scripts?

PR #144, opened 2017-06-02
Separate task state updates from worker updates

buchanae:
This move the task state updates to the task runner, separating it from the worker state service, which allows task runners to exist on their own without a worker service running. This can be useful for development and testing, and also allows more options for integrating with AWS Batch, HPC clusters, and more.

This currently passes the end-to-end tests, but I need to rewrite some of the other worker tests.

The majority of the changes are in worker/worker.go and worker/runner.go

I also added in some better error handling to the runner. It now catches panics, and returns tes.State_SYSTEM_ERROR where appropriate (basically every error that not from docker run)


worker/rpc_task.go
```diff
@@ -1,22 +1,56 @@
...
+	return &RPCTask{client, taskID, conf.UpdateTimeout}, nil
+}
+
+func (r *RPCTask) Task() (*tes.Task, error) {
+	return r.client.GetTask(context.Background(), &tes.GetTaskRequest{
+		Id:   r.taskID,
+		View: tes.TaskView_BASIC,
+	})
+}
+
```
adamstruck: Shouldn't Task() request the full view and State() the basic view?

buchanae: The worker probably doesn't need the full view, since that would just give it extra logs and add network overhead. The default is the minimal view, which is all `State()` needs.

adamstruck: Isn't this the Task method the runner is calling in runner.go (line 65) to get its Task?

buchanae: Oh, I guess the full view includes task inputs contents, but the basic doesn't? If so, it should be the full view.

buchanae:
Sorry, this has turned into a massive changeset. Changing the worker's task updates had a big effect on the test code, so many tests need to be moved/rewritten. On the upside, I was able to move most (or all) end-to-end tests into their proper home, remove the duplicated end-to-end test servers, and simplify many of the tests.

buchanae:
Note that in order to replace the GCE end-to-end tests, I needed to make a mock GCE scheduler backend. That required refactoring the scheduler's plugin code (and hopefully improving it) and moving the GCE metadata config code.

.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: go
+group: deprecated-2017Q2
```
adamstruck: can you remove this?

cmd/gce/gce.go
```diff
@@ -29,140 +24,30 @@ func init() {
...
-	// Auto detect the server's host name when it's not already set.
-	// This makes server deployment and configuration a bit easier.
-	// TODO will this work across zones?
-	if conf.HostName == defaultHostName && meta.Instance.Hostname != "" {
-		conf.HostName = meta.Instance.Hostname
-	}
 
-	return conf, nil
-}
+		logger.Configure(conf.Logger)
```
adamstruck: This should not be called until you know if your running a worker or server process. That way you know which Logger to load from the config.

worker/rpc_task.go
```diff
@@ -122,11 +159,47 @@ func (r *RPCTask) AppendExecutorStdout(i int, s string) {
...
+	return err
+}
+
+type taskClient struct {
+	tes.TaskServiceClient
+	pbf.SchedulerServiceClient
+}
+
+func newTaskClient(conf config.Worker) (*taskClient, error) {
+	conn, err := grpc.Dial(conf.ServerAddress, grpc.WithInsecure())
```
adamstruck: Should we have a timeout on this and/or should this block until the connection is made?

worker/runner_test.go
```diff
@@ -0,0 +1,41 @@
...
+		}
+		r.Run(context.Background())
+	*/
+
+	// Expect logging endpoints to be called
+	// Expect task to be set to running
+	// Expect task to be set to complete
+}
+
+func TestCancelContext(t *testing.T) {
```
adamstruck: I think these should either be defined or deleted

scheduler/pbs/sched.go
```diff
@@ -11,26 +11,22 @@ import (
...
-var Plugin = &scheduler.BackendPlugin{
-	Name:   "pbs",
-	Create: NewBackend,
-}
-
 // NewBackend returns a new PBS Backend instance.
 func NewBackend(conf config.Config) (scheduler.Backend, error) {
-	b := scheduler.Backend(&Backend{
-		name:     "pbs",
+	return &Backend{
```
adamstruck: having the 'return' here is an antipattern compared to the other backends

server/scheduler_service.go
```diff
@@ -135,6 +124,15 @@ func (taskBolt *TaskBolt) AssignTask(t *tes.Task, w *pbf.Worker) error {
 	})
 }
 
+// UpdateTaskState updates a task's state in the database.
+func (taskBolt *TaskBolt) UpdateTaskState(ctx context.Context, req *pbf.UpdateTaskStateRequest) (*pbf.UpdateTaskStateResponse, error) {
+	log.Debug("Update task STATE", req)
```
adamstruck: Is this needed?

tests/e2e/s3/s3_test.go
```diff
@@ -10,12 +10,15 @@ import (
 )
 
 var log = logger.New("e2e-s3")
-var fun *e2e.Funnel
+var fun = e2e.NewFunnel(e2e.DefaultConfig())
```
adamstruck: why not have fun and `fun.WithLocalBackend()` in `TestMain`?

tests/e2e/webdash_test.go
```diff
@@ -1,4 +1,4 @@
-package tests
+package e2e
```
adamstruck: Since all the tests seems to have moved into e2e does it still make sense to have the e2e subdirectory and package?

buchanae: Probably not. I'd like to do file moving in a PR where that's all we do.

worker/runner.go
```diff
@@ -184,106 +237,91 @@ func (r *taskRunner) fixLinks(basepath string) {
...
+					f()
+				}
+			}
 		}
-	}
+	}()
+	return taskctx
+}
+
+// NoopRunner is useful during testing for creating a worker with a Runner
```
adamstruck: I think NoopRunner and DefaultRunner should be in different files.

worker/runset.go
```diff
@@ -0,0 +1,67 @@
...
+		r.runners = make(map[string]context.CancelFunc)
+	}
+
+	// If there's already a runner for the given task ID,
+	// do nothing.
+	if _, ok := r.runners[id]; ok {
+		return
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
adamstruck: I thought we discussed this context being inherited from the worker

buchanae: Are you seeing an outdated file? This line is gone:
https://github.com/buchanae/funnel/blob/direct-to-task/worker/runset.go#L33

PR #145, opened 2017-06-02
Pagination for ListTasks

buchanae:
Needs tests, but seems to work with curl.

I made the minimum page size 50. Thoughts?


buchanae:
Found a bug, tests are good!

buchanae:
I updated the web dashboard to pull all ListTasks pages. The terminal dashboard requires that pages be added to the HTTP client first.

PR #150, opened 2017-06-07
Use TestMain instead of init

buchanae:
This makes the e2e test server more reusable by moving the `init()` code to `TestMain()`. The S3 tests also had a TestMain, so that moved into a subpackage to remove that conflict. Also cleaned up an unwanted debug log line.

### DIFF
--- a/tests/e2e/basic_test.go
+++ b/tests/e2e/basic_test.go
@@ -1,10 +1,19 @@
 package e2e
 
 import (
+	"github.com/ohsu-comp-bio/funnel/logger"
 	"github.com/ohsu-comp-bio/funnel/proto/tes"
+	"os"
 	"testing"
 	"time"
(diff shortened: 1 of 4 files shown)


tests/e2e/basic_test.go
```diff
@@ -1,11 +1,21 @@
...
+	"github.com/ohsu-comp-bio/funnel/logger"
 	"github.com/ohsu-comp-bio/funnel/proto/tes"
+	"os"
 	"testing"
 	"time"
 )
 
+func TestMain(m *testing.M) {
+	logger.ForceColors()
+	log.Debug("STARTGIN")
```
adamstruck: typo

buchanae: Fixed, thanks.

PR #153, opened 2017-06-08
Slurm, SGE, PBS/Torque Backends

adamstruck:
Added support for:

* Slurm
* SGE
* PBS/Torque

These backend are primarily controlled by their submit file templates that are provided in the configuration. 

Also added disk size detection (based on the configured `WorkDir` of the worker. 

Also added docker based testing of these backends.  These tests are run if the `-run-test` flag is passed to `go test`.

```
go test -v -timeout 120s ./tests/e2e/slurm -run-test
go test -v -timeout 120s ./tests/e2e/gridengine -run-test
go test -v -timeout 120s ./tests/e2e/htcondor -run-test
go test -v -timeout 120s ./tests/e2e/pbs -run-test
```


scheduler/scheduler.go
```diff
@@ -82,15 +82,22 @@ func (s *Scheduler) Schedule(ctx context.Context) error {
...
 		if offer != nil {
 			log.Info("Assigning task to worker",
 				"taskID", task.Id,
 				"workerID", offer.Worker.Id,
 			)
-			s.db.AssignTask(task, offer.Worker)
+			err = s.db.AssignTask(task, offer.Worker)
+			if err != nil {
+				log.Error("Error in AssignTask", err)
+				// return err
```
buchanae: What's correct here?

server/scheduler_service.go
```diff
@@ -42,6 +42,7 @@ func updateWorker(tx *bolt.Tx, req *pbf.Worker) error {
 	// Get worker
 	worker := getWorker(tx, req.Id)
 
+	// TODO fix concurrency bug
```
buchanae: Added #161 for this. I think you can remove this TODO.

tests/e2e/utils.go
```diff
@@ -51,17 +59,18 @@ func NewFunnel() *Funnel {
...
+	conf.ScheduleRate = time.Millisecond * 700
+	conf.Worker = config.WorkerInheritConfigVals(conf)
 
 	storageDir, _ := ioutil.TempDir("./test_tmp", "funnel-test-storage-")
 	wd, _ := os.Getwd()
 
 	conf.Storage = config.StorageConfig{
 		Local: config.LocalStorage{
-			AllowedDirs: []string{storageDir, wd},
+			AllowedDirs: []string{"./test_tmp", wd},
```
buchanae: Is this necessary? Does the server need access to storage outside of the storageDir?

tests/e2e/utils.go
```diff
@@ -72,34 +81,64 @@ func NewFunnel() *Funnel {
...
+
+// StartServer starts the server
+func (f *Funnel) StartServer() {
+	go server.Run(f.Conf)
+	f.pollForServerStart()
+	f.addRPCClient()
+}
+
+func (f *Funnel) pollForServerStart() {
+	timeout := make(chan bool, 1)
```
buchanae: Would `time.NewTimer()` be simpler?

tests/e2e/utils.go
```diff
@@ -72,34 +81,64 @@ func NewFunnel() *Funnel {
...
+	go func() {
+		time.Sleep(30 * time.Second)
+		timeout <- true
+	}()
+
+	srvRunning := make(chan bool, 1)
+	go func() {
+		for {
+			_, err := f.HTTP.GetServiceInfo()
```
buchanae: Don't you want a ticker here?

buchanae: Or even, use the RPC client and set it to retry, add a context with a timeout, and possibly you can collapse all this into a single line.

website/content/docs/backends/slurm.md
```diff
@@ -0,0 +1,48 @@
...
+to `sbatch`.
+
+The Funnel server process needs to run on the same machine as the SLURM master.  
+Configure Funnel to use SLURM by including the following config:
+
+```YAML
+Scheduler: slurm
+Backends:
+    SLURM: |
+        #!/bin/bash
```
buchanae: If it's easy to import these from a file in Hugo, it will prevent them from getting out of sync. It's very easy to use go-bindata to compile them into the binary from text files.

worker/util.go
```diff
@@ -100,3 +113,10 @@ func detectResources(conf config.Resources) config.Resources {
 // that doesn't do anything.
 func NoopTaskRunner(l TaskControl, c config.Worker, j *pbf.TaskWrapper) {
 }
+
+const (
+	b  = 1
+	kb = 1024 * b
+	mb = 1024 * kb
+	gb = 1024 * mb
```
buchanae: These are gibibytes (1024). Let's stick to SI units, which means these should be multiples of 1000. We might need to clarify this point in the TES docs as well.

worker/util.go
```diff
@@ -88,9 +98,12 @@ func detectResources(conf config.Resources) config.Resources {
...
 
-	if conf.RamGb == 0.0 {
-		res.RamGb = float64(vmeminfo.Total) /
-			float64(1024) / float64(1024) / float64(1024)
+	if conf.Resources.RamGb == 0.0 {
+		res.RamGb = float64(vmeminfo.Total) / float64(gb)
+	}
+
+	if conf.Resources.DiskGb == 0.0 {
+		res.DiskGb = float64(diskinfo.Free) / float64(gb)
```
buchanae: There was once a panic here, right? These should be error checked.

worker/util.go
buchanae: Oh, I was forgetting the context. The unit should match whatever gopsutil is returning. Also, "b", "kb", and "mb" aren't used anywhere, right?

worker/util.go
buchanae: Oh, I missed the recover above. Is nil-pointer the only panic we're catching? Because that's better done by normal checks. If there are potentially other panics because of syscall, then `recover()` makes sense.

worker/util.go
buchanae: I should think before I write. Since gopsutil is returning bytes, I'm pretty sure the correct number to use is 1000.

PR #154, opened 2017-06-12
Logger: config refactor and json for non-tty

buchanae:
The goal here was to allow future code more flexibility over logging. Previous to this PR, the global logger controlled the config of all child loggers. In the AWS Batch code, we might want a separate logger with a separate config. 

This also makes the default non-TTY output JSON formatted. JSON formatting should allow us to easily write utilities for dealing with log files (re-colorize, filter, etc). On the other hand, it's not very readable if you're just looking at the raw text. I think we agreed we're not 100% sure which is the best non-TTY format, but we're giving JSON a trial period.

This also just cleans up the code a bit, since a lot of it was copy-pasted from other projects.


cmd/worker/worker.go
```diff
@@ -41,14 +41,14 @@ func init() {
...
-	flags.StringVar(&flagConf.LogPath, "log-path", flagConf.LogLevel, "File path to write logs to")
+	flags.StringVar(&flagConf.Logger.Level, "log-level", flagConf.Logger.Level, "Level of logging")
+	flags.StringVar(&flagConf.Logger.OutputFile, "log-path", flagConf.Logger.OutputFile, "File path to write logs to")
 }
 
 // Run runs a worker with the given config, blocking until the worker exits.
 func Run(conf config.Config) error {
 
-	logutils.Configure(conf)
+	logger.Configure(conf.Logger)
```
adamstruck: pass conf.Worker.Logger

config/config.go
```diff
@@ -154,8 +148,7 @@ func WorkerInheritConfigVals(c Config) Worker {
 	}
 	c.Worker.Storage = c.Storage
 	c.Worker.WorkDir = c.WorkDir
-	c.Worker.LogLevel = c.LogLevel
-	c.Worker.TimestampLogs = c.TimestampLogs
+	c.Worker.Logger = c.Logger
```
adamstruck: Lets remove inheritance of `Logger` and `WorkDir`

PR #156, opened 2017-06-16
AWS batch proxy

buchanae:
TODO:
- [ ] Ensure executor logs are trimmed to appropriate size
- [ ] Decide how to handle finished jobs being deleted after 24 hours. Documentation?
    - `The job state for SUCCEEDED jobs is persisted in AWS Batch for 24 hours.`
    - `The job state for FAILED jobs is persisted in AWS Batch for 24 hours.`
- [ ] API calls could use context with timout
- [ ] Proxy does not currently return SYSTEM_ERROR. Determine if there are any cases where this error is more appropriate.
- [ ] ListTasks doesn't match Batch very well. Currently ListTasks makes multiple calls to AWS Batch ListJobs, which is slow and inefficient. For similar reasons, ListTasks does not support pagination.
    - If multiple job queues were used, this would be even more complex
    - Also, the AWS batch job IDs aren't consistent with other Funnel Task IDs: funnel IDs are sortable, while AWS is using generic UUIDs.
- [ ] Allow full configuration of AWS compute environment, e.g. Ec2KeyPair, DesiredVCPUs, etc.
- [ ] Auto initialization/detection of configuration, includes detecting/determining security groups, subnets, etc.
- [ ] Task views (Basic, full) are not supported.
- [ ] Config validation
    - job queue compute environment name should match compute env name
- [ ] This requires a published docker container, so we should make sure that container is updated by our PRs and builds
- [ ] Deployment script should handle the case where the entity already exists, gracefully. Currently for compute env. and job queue it's an error. How would the user update those?
- [x] The web dashboard is not currently served.
- [ ] `funnel aws runtask` should call `tes.Validate()` before running
- [x] `funnel aws runtask` should set up storage configuration
- [x] `funnel aws runtask` should return the exit code of the task runner.
- [x] Test with S3 auth. **failed**
  - [ ] Not simple to get S3 key/secret from the environment? And IAM roles are the recommend way.
- [ ] Look into the cost of storing logs in CloudWatch. It might be better to store some or all of the logs in S3.
  - https://aws.amazon.com/cloudwatch/pricing/
  - $0.50 per GB ingested**
  - $0.03 per GB archived per month***
  - S3: Upon sign-up, new AWS customers receive 5 GB of Amazon S3 storage in the Standard Storage class, 20,000 Get Requests, 2,000 Put Requests, and 15 GB of data transfer out each month for one year.
    - First 50 TB / month: $0.023 per GB
- [ ] Find a good home for this code in the codebase. It's a different model than the server/scheduler/worker stuff, so not sure where it fits right now.
- [ ] Web dashboard shows state "UNKNOWN" in list view
- [ ] A SYSTEM_ERROR state from the runner is not reflected by GetTask
- [ ] Discuss how S3 endpoints are configured from a task message.


adamstruck:
closing in favor of #246

PR #157, opened 2017-06-19
Basic auth (password only) for API

buchanae:
This is pretty quick and hacky, but does work. Still needs thought/work though.

TODO
- [x] TLS between clients/server and server/worker/scheduler (pushed to future task)
- [x] Tests
- [x] Docs, especially about storing a password in a plaintext config file.


website/static/funnel-config-examples/config.go
```diff
@@ -0,0 +1,280 @@
+package config
```
adamstruck: this shouldn't be in here

website/static/funnel-config-examples/config_test.go
```diff
@@ -0,0 +1,25 @@
+package config
```
adamstruck: also doesn't belong here

server/scheduler_service.go
```diff
@@ -62,6 +62,8 @@ func updateWorker(tx *bolt.Tx, req *pbf.Worker) error {
 		}
 	}
 
+	log.Debug("TASKS", req.TaskIds)
```
adamstruck: remove this

PR #167, opened 2017-07-02
Start separate server per test in e2e

buchanae:
direct-to-task should be merged first.

The e2e test logs are too difficult to debug on failure. Whatever benefit might be gained in faster build times or more robust testing is easily lost to frustration and time spent puzzling over bugs.

This starts a separate server per test, which should help cut down the logs printed on test failure.


buchanae:
Hrm, I guess the logs for all tests are still dumped on a single failure? I'll need to find a different approach...

PR #174, opened 2017-07-17
termdash: fix panic, reverse order, full task IDs, cleanup

adamstruck:
Rather than panicking, the dashboard will render and an error will appear in the header portion of the UI.

### DIFF
--- a/cmd/termdash/compact/main.go
+++ b/cmd/termdash/compact/main.go
@@ -22,14 +22,9 @@ type Compact struct {
 }
 
 func NewCompact(t *tes.Task) *Compact {
-	// truncate task id
-	id := t.Id
-	if len(id) > 12 {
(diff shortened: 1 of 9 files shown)


buchanae:
Looks like this also:
- remove the truncation of task IDs
- Reverses the task sort order, so most recent tasks are on top
- Has some minor code cleanup

Did I miss anything? +1 to all of those, just wanted to note it for the record.

In the future, I'd recommend you avoid using panic/recover, but I'm not going to be too picky here.

adamstruck:
Yeah all those things as well.  I had this branch sitting around and figured I might as well get some of it merged in. 

Yeah I would have preferred not to use panic/recover, but since this was mostly a port of ctop I just went with the patterns that were there.

Issue #172, opened 2017-07-07
`funnel task get` with invalid task ID returns message instead of 404


buchanae:
```
./bin/funnel task get foo123
{
	"id": "foo123",
	"logs": [
		{

		}
	]
}
```

Issue #171, opened 2017-07-07
Panic from `funnel task list`


buchanae:
Looks like my pagination change might have broken something:
```
funnel task list
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x48662bf]

goroutine 1 [running]:
github.com/ohsu-comp-bio/funnel/cmd/task.doList(0x4b2c9d8, 0x15, 0xc4200a8d20, 0x4b1b8cb, 0x4, 0x4b1b763)
	/Users/buchanae/go/src/github.com/ohsu-comp-bio/funnel/cmd/task/list.go:54 +0x12f
github.com/ohsu-comp-bio/funnel/cmd/task.glob..func4(0x50b6040, 0x50dcdc0, 0x0, 0x0, 0x0, 0x0)
	/Users/buchanae/go/src/github.com/ohsu-comp-bio/funnel/cmd/task/list.go:24 +0x3d
github.com/ohsu-comp-bio/funnel/vendor/github.com/spf13/cobra.(*Command).execute(0x50b6040, 0x50dcdc0, 0x0, 0x0, 0x50b6040, 0x50dcdc0)
	/Users/buchanae/go/src/github.com/ohsu-comp-bio/funnel/vendor/github.com/spf13/cobra/command.go:647 +0x3f8
github.com/ohsu-comp-bio/funnel/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0x50b4060, 0x0, 0x0, 0x50b4280)
	/Users/buchanae/go/src/github.com/ohsu-comp-bio/funnel/vendor/github.com/spf13/cobra/command.go:726 +0x339
github.com/ohsu-comp-bio/funnel/vendor/github.com/spf13/cobra.(*Command).Execute(0x50b4060, 0x50b4f40, 0xc420263f68)
	/Users/buchanae/go/src/github.com/ohsu-comp-bio/funnel/vendor/github.com/spf13/cobra/command.go:685 +0x2b
main.main()
	/Users/buchanae/go/src/github.com/ohsu-comp-bio/funnel/main.go:11 +0x31
```

I'll make a fix that includes a test for this command

Issue #169, opened 2017-07-05
GetTask + basic view incorrectly returns input "contents"


buchanae:
This should not return the "contents" field:

```
funnel task get b5eim3qrl6qt06ads780
{
	"id": "b5eim3qrl6qt06ads780",
	"state": "COMPLETE",
	"name": "Funnel run: sh -c \"echo $IN\"",
	"inputs": [
		{
			"name": "IN",
			"path": "/opt/funnel/inputs/Users/buchanae/go/src/github.com/ohsu-comp-bio/funnel/README.md",
			"contents": "[![Build Status](https://travis-ci.org/ohsu-comp-bio/funnel.svg?branch=master)](https://travis-ci.org/ohsu-comp-bio/funnel)\n[![Gitter](https://badges.gitter.im/ohsu-comp-bio/funnel.svg)](https://gitter.im/ohsu-comp-bio/funnel)\n[![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)\n[![Godoc](https://img.shields.io/badge/godoc-ref-blue.svg)](http://godoc.org/github.com/ohsu-comp-bio/funnel)\n\n\nFunnel (alpha)\n======\n\nFunnel is a server for executing tasks on a cluster. Given a task description, Funnel will find a worker to execute the task, download inputs, run a series of (Docker) containers, upload outputs, capture logs, and track the whole process.\n\nFunnel is an implementation of the [GA4GH Task Execution Schemas](https://github.com/ga4gh/task-execution-schemas), an effort to standardize the APIs used for task execution across many platforms.\n\nFunnel provides an API server, multiple storage backends (local FS, S3, Google Bucket, etc.), multiple compute backends (local, HTCondor, Google Cloud, etc.), and a web dashboard.\n"
		}
	],
	"resources": {

	},
	"executors": [
		{
			"imageName": "alpine",
			"cmd": [
				"sh",
				"-c",
				"echo $IN"
			],
			"workdir": "/opt/funnel",
			"stdout": "/opt/funnel/outputs/stdout-0",
			"stderr": "/opt/funnel/outputs/stderr-0",
			"environ": {
				"IN": "/opt/funnel/inputs/Users/buchanae/go/src/github.com/ohsu-comp-bio/funnel/README.md"
			}
		}
	],
	"logs": [
		{
			"logs": [
				{
					"startTime": "2017-07-05T11:08:17-07:00",
[...]

PR #175, opened 2017-07-17
ListTask command fix and GetTask endpoint fix

adamstruck:
* Fixed panic in `funnel task list`  that occurred when there were zero tasks in the DB.
* GetTask now returns 404 when a request is sent with an unknown task id

### DIFF
--- a/server/task_boltdb.go
+++ b/server/task_boltdb.go
@@ -194,13 +194,31 @@ func getTaskState(tx *bolt.Tx, id string) tes.State {
 	return tes.State(v)
 }
 
+type taskNotFoundError struct {
+	id string
+}
+
+func (e *taskNotFoundError) Error() string {
+	return fmt.Sprintf("no task found for id: %s", e.id)
+}
+
+func loadMinimalTaskView(tx *bolt.Tx, id string, task *tes.Task) error {
+	b := tx.Bucket(TaskBucket).Get([]byte(id))
+	if b == nil {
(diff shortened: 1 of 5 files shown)


adamstruck:
Addresses #171 and #172

server/task_boltdb.go
```diff
@@ -221,32 +225,39 @@ func loadTaskLogs(tx *bolt.Tx, task *tes.Task) {
...
-	err := taskBolt.db.View(func(tx *bolt.Tx) error {
-		task = getTaskView(tx, req.Id, req.View)
-		return nil
+	var err error
+	err = taskBolt.db.View(func(tx *bolt.Tx) error {
+		task, err = getTaskView(tx, req.Id, req.View)
+		return err
 	})
-	return task, err
+	if err != nil {
```
buchanae: This will mark every error as a missing task, which will be misleading for an error at some point. You can get around this by returning an errTaskNotFound  value from loadBasicTaskView, instead of a generic fmt.Errorf.

Also, I may be reading the code wrong, but I think this will still incorrectly return a task message for a non-existant task when using the MINIMAL view.

buchanae:
Also, tests plzzzz. kthxbye.

server/task_boltdb.go
adamstruck: Right now that is the only type of error that can occur, but that may not always be true. Good points.

adamstruck:
added a fix for #169

PR #176, opened 2017-07-19
updated version in makefile and upload-release cmd

adamstruck:


Makefile
```diff
@@ -114,19 +114,12 @@ upload-release:
...
 	@cp build/bin/* build/release/
 	@cp build/funnel-gce-image-installer build/release
-	@for GOOS in darwin linux; do \
-		for GOARCH in amd64; do \
-			GOOS=$$GOOS GOARCH=$$GOARCH \
-				tar -C build/release -czvf build/release/funnel-$$GOOS-$$GOARCH.tar.gz funnel-$$GOOS-$$GOARCH; \
-				rm build/release/funnel-$$GOOS-$$GOARCH; \
-		done; \
-	done
-	#github-release release \
```
buchanae: Is this being removed because it's exactly the same as cross-compile?

adamstruck: Yes and make gce-installer before this which calls cross compile.

PR #185, opened 2017-07-27
WIP: SSL

buchanae:
This is what I have for SSL so far. I need to switch to other tasks, so putting this up to share.

There's more to do:
- [ ] tests
- [ ] the CLI clients need to allow config to set up self-signed certs, and/or...
- [ ] the code should load the certs installed on the OS using `x509.NewCertPool` (or something)

Note that `cmd/server/server.go` changed to load the server before the scheduler, which fixed connection errors from grpc being logged.

This has been helpful:
https://bbengfort.github.io/programmer/2017/03/03/secure-grpc.html

### DIFF
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -70,6 +70,19 @@ func Run(conf config.Config) error {
 		return err
 	}
 
+	srv := server.DefaultServer(db, conf)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Start server
+	var srverr error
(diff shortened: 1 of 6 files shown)


buchanae:
Oh, also, "localhost" is hard-coded here:
https://github.com/ohsu-comp-bio/funnel/pull/185/files#diff-e95b5efe037d7dad38e3a4f28506b768R22

buchanae:
Super old and not currently active. I'll dig this PR out if I need it in the future.

PR #189, opened 2017-08-03
Swift storage backend

buchanae:
Work in progress

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -266,6 +266,7 @@ type StorageConfig struct {
 	Local LocalStorage
 	S3    S3Storage
 	GS    []GSStorage
+	Swift SwiftStorage
 }
 
 // LocalStorage describes the directories Funnel can read from and write to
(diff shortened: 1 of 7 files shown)


buchanae:
Needs tests, but otherwise seems to work in my basic tests on exastack.

Need to figure out how to test something like this (and Google Storage, AWS S3, etc).

buchanae:
zomg, it builds.

buchanae:
The swift backend was not failing when it failed to access files owned by root. So there's a bug in there somewhere.

adamstruck:
Can you also add a test for get/put on a directory?

config/config.go
```diff
@@ -255,6 +258,22 @@ func (l S3Storage) Valid() bool {
...
 
+// SwiftStorage configures OpenStack Swift object storage.
+type SwiftStorage struct {
+	UserName   string
+	Password   string
+	AuthURL    string
+	TenantName string
+	TenantID   string
+	RegionName string
+}
```
adamstruck: Could you add an option for these to be loaded from environmental variables?

adamstruck:
- [ ] load swift config vars from environment
- [ ] squash commits
- [ ] rebase

PR #194, opened 2017-08-11
Add context to worker logging

buchanae:
This adds handling of context values in logging arguments, and adds contexts to log calls in the worker and storage. This helps ensure that values such as taskID and workerID are included in log calls from components such as DockerCmd.


buchanae:
Closed in favor of #201

PR #201, opened 2017-08-25
Axe the global logger in the worker package

buchanae:
Here's a prototype of what it might look like to start removing the global logger. This would make #194 obsolete.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"fmt"
 	"github.com/ghodss/yaml"
 	"github.com/ohsu-comp-bio/funnel/logger"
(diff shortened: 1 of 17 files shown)


buchanae:
Another consideration for logging: As I've worked on smc-het, I've found it would be useful to turn on debug level for a single task. For example, some task may be exposing a bug in funnel, and the cluster is currently at the INFO level. I'd like to turn on debugging in order to collect better information to understand what is going on. Currently, I'd need to update some crazy ansible config and redeploy the entire cluster just to get debug logs, or alternatively I'd have to set up a development cluster and try to reproduce the bug there. Both are pretty heavy options.

buchanae:
This is also better handled by #235. There are a few things left over, such as capturing logs from storage, which I'll revisit in a new PR.

PR #205, opened 2017-08-30
FUNNEL_SERVER env var for CLI

buchanae:
This is my minimal effort, quick fix for avoiding writing `funnel -S http://some-funnel-server:8000` all the time.

### DIFF
--- a/cmd/run/flags.go
+++ b/cmd/run/flags.go
@@ -138,6 +138,7 @@ func defaultVals(vals *flagVals) {
 		vals.name = "Funnel run: " + vals.execs[0].cmd
 	}
 
+	vals.server = os.Getenv("FUNNEL_SERVER")
 	if vals.server == "" {
 		vals.server = "http://localhost:8000"
 	}
--- a/cmd/task/task.go
+++ b/cmd/task/task.go
@@ -22,7 +22,7 @@ func init() {
 		tesServer = "http://localhost:8000"
 	}
 
-	Cmd.PersistentFlags().StringVarP(&tesServer, "server", "S", tesServer, "")
+	Cmd.PersistentFlags().StringVarP(&tesServer, "server", "S", tesServer, `may be set by FUNNEL_SERVER env var`)
 	Cmd.AddCommand(listCmd)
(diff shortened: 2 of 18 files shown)


adamstruck:
Please document this new feature in the cli help message

buchanae:
The last commit added a note to the CLI flag help. It's not the nicest solution, but I'm not sure where else to document it yet.

buchanae:
I started experimenting with testing using shell scripts, which are in `tests/e2e/sh`

Benefits:
- more emphasis on paying attention to cli output and actual JSON task messages. I started noticing things about task messages I never noticed before. In fact, I think it revealed a few bugs.
  - in OutputFileLog.sizeBytes in JSON is a string
  - `task list` output is not pretty printed
  - `run --contents` doesn't set the input correctly
  - there are more. I'll deal fixing these in other issues/PRs
- pretty simple test harness, see `tests/e2e/sh/sh_test.go`. Tests are a shell script and expected stdout/err output. Substantially more simple than the existing e2e tests.
- would allow for testing CLI help messages
- moves a little closer towards having a set of scripts we could provide as TES conformance tests
- builds a corpus of examples we could point people to

Caveats:
- I stripped id/startTime/endTime/hostIp lines from the output because those are non-deterministic
- Will need some clever way to handle file paths in task messages
- Depends on the funnel binary you want to test being already built and in the PATH. Not perfect, but likely to always be true for us.

In the last commit, I've avoided dealing with these caveats and included only a subset of the e2e tests which were simple to implement. The rest of the tests are [here](https://github.com/buchanae/funnel/tree/sh-tests/tests/e2e/sh)

These tests all depend on the FUNNEL_SERVER env var, which is why I'm including them in this PR.

adamstruck:
I think I'd prefer most of `tests/e2e/sh` to go into a different PR since most of them are reimplementing existing tests. I do like this approach though - I think it makes the tests clearer and more concise. 

In this PR I'd prefer to see only tests relevant to the code changes. Test that the env var is used, test that the flag overwrites that value, etc.

buchanae:
Replaced by #256

PR #209, opened 2017-09-03
Stateless workers design

buchanae:
This starts a `designs` folder to keep design documents in github instead of google docs, starting with the design for scheduler/stateless workers refactor.

@kellrott wanted to move designs into github, so he can provide more detail about the motivation for that if needed. Some benefits I can think of:

* closer to having everything in once place.
* github usernames instead of google accounts, means clearer relationship to the code.
* follows the pull request model for proposing changes and ideas.
* more activity in the repo, which might be noticeable.
* even if we use Google Docs, we still need a page linking out to all those Docs.

Why I didn't use the wiki:

* wiki can't have a pull request, which will be useful for discussion and comments.
* wiki lacks good organization tools.
* wiki is not part of the codebase and version history, like everything else is.
* wiki requires additional user permission admin, while pull requests are open to anyone.

That said, I'm not strongly opinionated on any of this. I don't care where we put designs, as long as we do it consistently. I think putting it in the codebase works fine and other projects have done this successfully.

### DIFF
--- a/designs/01-stateless-workers.md
+++ b/designs/01-stateless-workers.md
@@ -1,0 +1,107 @@
+# Stateless workers
+
+Funnel should not manage persistent workers itself.
+
+A “persistent worker” is a long-lived service/daemon worker process,
+which processes multiple tasks tasks over time.
+These long-lived processes come with additional complexity:
+
+* Restarting a worker takes care.
+* If a worker is restarted, it will lose track of all its running tasks.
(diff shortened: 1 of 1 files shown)


buchanae:
One downside of using github markdown is that small edits become a commit, which gets a little tedious and spammy.

buchanae:
Don't hear a strong push against google docs, which seems to be preferred by the funnel devs.

PR #215, opened 2017-09-05
Split TaskService into TaskReader + TaskWriter

buchanae:
This PR attempts to clean up the TaskService interface by splitting it into two new/renamed interfaces: TaskReader and TaskWriter.

The main motivation here is to allow me to more easily see all the logs related to a task worker/runner in one place. Currently, I get part of the info from GetTask, and the other part from Kibana, which makes debugging failed tasks more difficult.

By separating these two, I can easily write a second TaskWriter which emits log messages via Funnel's logger package. This writer can be composed with the RPC-based writer, to allow both updates and log messages.

An important note is that `TaskService.SetState` moves to `TaskWriter.State`, which was the main reason I ended up renaming and reorganizing the entire interfaces (`TaskLogger.State` would have conflicted with `TaskService.State`). Also, the names are much cleaner now, I think.

Based on the worker logging cleanup branch from #201

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"fmt"
 	"github.com/ghodss/yaml"
 	"github.com/ohsu-comp-bio/funnel/logger"
 	os_servers "github.com/rackspace/gophercloud/openstack/compute/v2/servers"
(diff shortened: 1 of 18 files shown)


buchanae:
Better handled by #235

PR #217, opened 2017-09-05
Task multi writer

buchanae:
This PR includes a TaskWriter that both writes to RPC and logs via Funnel logging.


buchanae:
Better handled by #235

PR #234, opened 2017-09-14
Amazon DynamoDB

adamstruck:
Added Amazon DynamoDB server backend

DynamoDB TODO:
- [x] handle updates to `Stdout` and `Stderr` in `ExecutorLog`
- [x] enforce state transition rules
- [x] return tasks ordered by creation time from `ListTasks`
- [x] ~~enable configuration of autoscaling~~
- [x] pass context to API calls
- [x] stdout and stderr is stored without funnel imposed limits #247
- [x] tests
- [x] documentation #248


adamstruck:
DynamoDB TODO:
- [ ] handle updates to `Stdout` and `Stderr` in `ExecutorLog`
- [ ] enforce state transition rules
- [ ] return tasks ordered by creation time from `ListTasks`
- [ ] enable autoscaling with config?

adamstruck:
Returning tasks in order of creation ListTasks introduces limits in DynamoDB. 
*  For every distinct partition key value, the total sizes of all table and index items cannot exceed 10 GB. 

If we use a fixed partition key then we also have limits: 
* A single partition can support a maximum of 3,000 read capacity units or 1,000 write capacity units.

adamstruck:
closes #231

adamstruck:
**Subsequent Allocation of Partitions**

A single partition can hold approximately 10 GB of data, and can support a maximum of 3,000 read capacity units or 1,000 write capacity units.

If necessary, DynamoDB can allocate additional partitions to your table by splitting an existing partition. Suppose that one of a table's partitions (P) exceeded its limit for storage (10 GB). In this case, DynamoDB would split the partition as follows:
1.  Allocate two new partitions (P1 and P2).
2. Distribute the data from P evenly across P1 and P2.
3. Deallocate P from the table.

PR #235, opened 2017-09-17
Events

buchanae:
@adamstruck I pulled your events code into this PR

After using it for the os/exec worker branch, I made some changes:

- dropped the oneof
- dropped the proto Struct type for log fields
- moved proto file to events package
- added some additional writer/logger/gen helpers
- remove the "executor" prefix from stdout/err, hostip, etc. because it seemed overly verbose. start/end time still have the prefix.

I found the oneof to create a bunch of stuttering fields which seemed unnecessary. e.g.
```go
if e, ok := x.Event.(*events.Event_Stdout); ok {
	if e.Stdout.Stdout != "a" {
		t.Error("expected 'a' but got %s", e.Stdout.Stdout)
	}
}
```
vs
```go
if x.Type == events.Type_STDOUT && x.Stdout != "a" {
  t.Error("expected 'a' but got '%s'", x.Stdout)
}
```

I also found the Struct type difficult to work with. For logging, are we ok with a simple `map[string]string`? That covers most info I can think of. Is there a way to marshal/unmarshal between an arbitrary Go struct to a protobuf Struct? Having to write a type switch for all the Value types seems overkill, https://godoc.org/github.com/golang/protobuf/ptypes/struct#Value. Maybe I just need an example of the right way to do it.


buchanae:
The latest commit removes the `tailer`. This was a custom buffer type which started to get in the way because it didn't implement io.Reader/Writer interfaces. It also introduces strange behavior, where there may be a gap in the logs written to the event stream during network error conditions.

TODO:
- [ ] add back log event buffer and rate limiter

adamstruck:
I still prefer using oneof for the events. I'm fairly certain you can avoid the situations you describe and I think there are other benefits to the original approach.

PR #236, opened 2017-09-18
Create scheduler.md

buchanae:
Trying out what it's like to use a markdown file + github editor + pull requests to evolve a design document. Also, I'd like to capture some notes about the scheduler design, since it gets revisited often.


buchanae:
Adding a new commit to the PR via the editor seems to work ok.

PR #237, opened 2017-09-18
Clean dead nodes on signal and timeout

buchanae:
This improves our handling of dead nodes:
- `node run` now watches for SIGINT and SIGTERM (canceled process) and cancels the node's context. The node the tells the server it's gone, which deletes it from the DB.
- Adds a check for nodes that have been dead for more than 5 minutes, at which point they are deleted.

Hopefully, this will help us handle preempted tasks in the future.


adamstruck:
Needs a rebase, but otherwise looks like a nice improvement.

PR #241, opened 2017-09-20
Task restart

buchanae:
Here's a rough version of task restarts. We should wait until it gets some testing.

https://github.com/buchanae/funnel/commit/754a052abd974c76414fba2791c4ab8847bc7d4c fixes the server start, which can be pulled into a separate PR. Needs a test.

### DIFF
--- a/cmd/gce/gce.go
+++ b/cmd/gce/gce.go
@@ -1,6 +1,7 @@
 package gce
 
 import (
+	"context"
 	"fmt"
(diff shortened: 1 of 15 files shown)


buchanae:
- [ ] This is the last block that needs to be included from this PR. https://github.com/ohsu-comp-bio/funnel/pull/241/files#diff-431b7c73e7ae668fa229624a40171b57R248

buchanae:
Handled #409

PR #245, opened 2017-09-28
Fix server panic on manual backend

buchanae:
Fix for server panic with manual backend. Also trying out a shell test for this fix.


buchanae:
Shell tests are too much trouble. Going to find a way to test this via Go + Cobra.

PR #246, opened 2017-09-28
AWS Batch

adamstruck:
This PR adds batch as a compute backend. It also includes a cli utility for creating a compute environment, job queue, and job definition for use with this backend. 

TODO:
- [x] tests #272 
- [x] documentation


buchanae:
Just going to keep some rough notes here about the docs as I go through them:
- typo "reccommend"
- would help to have a bulleted list in the intro that describes what we're going to do. e.g. I wasn't sure if I'm supposed to create a Dynamo db first.

PR #254, opened 2017-10-04
Elastic sched

buchanae:
This adds an elasticsearch database backend for with APIs for TES, Events, and Nodes.

One major change was moving the scheduler's node update code out of the database, and simplifying the node database API. Databases are now just responsible for simple get/put/delete/list for nodes, with optimistic locking via the "Version" field (bolt had this already). Most of the node update code moved to the node.

Related changes:
- boltdb export, elasticsearch import commands (so I could migrate an existing database)
- added TaskCreated event to do the migration above
- merged in #245 to fix panics
- moved the scheduling/node update logic out of the database code
- the worker now gets a full task, instead of a task ID, because the node also needs the task, so it might as well get it and pass it to the node. This cleans up the TaskReader stuff a bit, and should fit better for a AWS Batch worker as well.
- pushing some config/util code around so that common RPC fields are shared more easily.
- fixing the GCE mock cycle that prevented gen-mocks from running.
- node.runSet is replaced with sync.Map, since I now need to iterate over the current tasks, deleting a bunch of code and using a core lib. instead.
- worker factory is gone. The node is passed a worker instead. Workers are expected to handle Run() being called with different tasks.

TODO
- [x] fix GCE tests and possibly rethink autoscaling there.
- [x] figure out switch statements for databases, compute backends, and task readers.
- [x] clean out debug lines
- [x] check for unused mocks
- [x] double-check task cancel, since the cancel signal changed.
- [ ] run HPC/GCE tests.
- [ ] get rid of SimpleClient
- [x] sorting is broken
- [ ] paging is broken
- [x] task views are broken
- [x] cancel is not implemented
- [ ] limiting stdout/err size is not implemented
- [x] fix node workdir config (for disk resource detection)
- [x] move code to fit more like bolt/dynamo
- [x] serialize enums as strings
- [x] write individual events into separate index
- [ ] elastic search event writer in the worker
- [x] write syslogs to a separate index


buchanae:
Split into multiple PRs: #245, #259, #260, #261

PR #256, opened 2017-10-09
Funnel server env var

buchanae:
@adamstruck want to get your thoughts on moving the commands to a function with hooks, so that we can more easily test the result of loading the command line + env.

I haven't written an example test that uses this yet, but essentially you'd do:
```go
func TestCmd(t *testing.T) {
  h := task.Hooks{
    Create: func() {
      // test code here
    }
  }
  cmd := task.NewCommand(h)
  cmd.SetArgs([]string{"create hello.json"})
}
```

This allows you to test that the server, view, config, etc. were properly loaded, but without actually needing to set up a server to connect to. I wrote this because I couldn't think of a way to test that the default TES server is "localhost:8000".


adamstruck:
Based on cobra's internal tests I think we can do it without refactoring the code to use hooks. Using `PreRun` to get the env var and set the server address should work and wouldnt require a server.

PR #261, opened 2017-10-11
Elasticsearch

buchanae:
More simple, updated version of #254 

Still a few outstanding issues from the checklist in #254 which I'll open as separate issues.


buchanae:
Closes #250

buchanae:
Latest commit closes #263

adamstruck:
I think we should either keep the `TaskGetter` or `TaskReader` interface. There is no need for both to exist.

PR #268, opened 2017-10-13
logging: remove global loggers

buchanae:
This removes all the global loggers. Turns out, this wasn't nearly as big a change as I expected. Most of the changes were things that should have been a returned error anyway.

There is work left to do:
- [x] account for all the removed Debug lines. Maybe a global debug logger is ok?
- [x] make e2e tests use a logger that wraps testing.T


compute/gce/sched.go
```diff
@@ -54,8 +49,6 @@ type Backend struct {
 
 // GetOffer returns an offer based on available Google Cloud VM node instances.
 func (s *Backend) GetOffer(j *tes.Task) *scheduler.Offer {
-	log.Debug("Running GCE backend")
```
buchanae: Put these in the scheduler

tests/e2e/slurm/slurm_test.go
```diff
@@ -8,13 +8,11 @@ import (
 	"testing"
 )
 
-var log = logger.New("e2e-slurm")
 var fun *e2e.Funnel
 var runTest = flag.Bool("run-test", false, "run e2e tests with dockerized scheduler")
+var log = logger.NewLogger("slurm", logger.DebugConfig())
```
adamstruck: This file should be consistent with gridengine, htcondor, etc.  In those files you changed `TestMain` to use `logger.Debug` rather than declaring a global logger.

tests/e2e/dynamodb/dynamodb_test.go
```diff
@@ -47,8 +44,7 @@ func TestMain(m *testing.M) {
 	defer deleteTables(c.Server.Databases.DynamoDB)
 	ok := checkTablesAreALive(c.Server.Databases.DynamoDB)
 	if !ok {
-		log.Error("Dynamodb tables were not active within timeout")
-		return
+		panic("Dynamodb tables were not active within timeout")
 	}
 
 	m.Run()
```
adamstruck: Shouldn't all of the tests in this file also use `setLogOutput`?

tests/e2e/utils.go
```diff
@@ -33,11 +33,20 @@ import (
...
 )
 
-var log = logger.New("e2e")
+var log = logger.NewLogger("e2e", testutils.LogConfig())
+
+func init() {
+	logger.ConfigureGRPC(1, log)
+}
+
+func setLogOutput(t *testing.T) {
```
adamstruck: This should be public so the other e2e tests can use it.

PR #281, opened 2017-10-27
worker: handle docker exit code 125 as syserr

buchanae:


adamstruck:
Not sure how I feel about this yet. 

Things to consider:
* It's possible to have the command in the container return 125. 
* Other exit codes with special meaning in docker (https://docs.docker.com/engine/reference/run/#exit-status).

buchanae:
Unlikely the container will return 125 and it's an improvement over thinking these failures are coming from the container. We can move to the docker client later to further improve this, maybe.

code 126 and 127 are both related to the command running inside the container, so should communicate that the executor command is wrong.

adamstruck:
OK. Add a test and I'll merge it.

buchanae:
I don't think there's an easy way to force docker to return a 125 error, since it's a daemon issue.

adamstruck:
An invalid flag would do it, but thinking about it there isn't a good way to make the worker do that either.

adamstruck:
if the executor references an image that does not exist then docker exits with 125

buchanae:
I don't have a good enough reason for this.

PR #283, opened 2017-10-31
Adding debug log messages

adamstruck:


buchanae:
This seems useful, but super spammy. I'd like to hide this behind a debug2 level or something, so that I can still debug my server without seeing every PutNode/GetNode request.

I also added this:
```
diff --git a/server/server.go b/server/server.go
index f49de6f..b1e5004 100644
--- a/server/server.go
+++ b/server/server.go
@@ -57,10 +57,16 @@ func newDebugInterceptor(log *logger.Logger) grpc.UnaryServerInterceptor {
        return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo,
                handler grpc.UnaryHandler) (interface{}, error) {
                log.Debug(
-                       "received request",
+                       "request: "+info.FullMethod,
                        "request", req,
                )
-               return handler(ctx, req)
+               resp, err := handler(ctx, req)
+               log.Debug(
+                       "responding: "+info.FullMethod,
+                       "resp", resp,
+                       "err", err,
+               )
+               return resp, err
        }
 }
```

adamstruck:
yeah I have been finding it a bit to verbose at times as well.

buchanae:
TODO: 
- [ ] add method name patch
- [ ] hide this behind debug2 log level, or some other way to opt in to this extremely verbose logging.

PR #291, opened 2017-11-02
OutputFileLog.path container path bug fix

### LINKED ISSUE #290: Output Log Path using host paths
Looking at the `logs.outputs.path` it looks like the path from the host VM is being recorded, rather then the docker container path. Thus it looks like `/mnt/funnel/b7to4ir3qt7gv3763k7g/var/run/cwl/gaf.json`

Example
```
"logs" : [
      {
         "outputs" : [
            {
               "path" : "/mnt/funnel/b7to4ir3qt7gv3763k7g/var/run/cwl/gaf.json",
               "size_bytes" : "70989749",
               "url" : "swift://biostream/job1/output/gaf.json"
            }
         ],
         "logs" : [
            {
               "host_ip" : "10.50.50.55",
               "start_time" : "2017-11-02T20:38:46.032906942Z",
               "end_time" : "2017-11-02T20:39:28.485524782Z"
            }
         ],
         "end_time" : "2017-11-02T20:39:30.815287726Z",
         "start_time" : "2017-11-02T20:38:44.187495026Z"
      }
   ],
```

adamstruck:
- deduplicated storage code for put
- report container paths in OutputFileLog

closes #290

### DIFF
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -13,6 +13,7 @@ import (
 	"github.com/ohsu-comp-bio/funnel/proto/tes"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 const (
@@ -26,7 +27,7 @@ const (
 // New storage backends must support this interface.
 type Backend interface {
 	Get(ctx context.Context, url string, path string, class tes.FileType) error
-	Put(ctx context.Context, url string, path string, class tes.FileType) ([]*tes.OutputFileLog, error)
+	PutFile(ctx context.Context, url string, path string) error
 	// Determines whether this backends supports the given request (url/path/class).
 	// A backend normally uses this to match the url prefix (e.g. "s3://")
 	// TODO would it be useful if this included the request type (Get/Put)?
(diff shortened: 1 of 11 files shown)


storage/storage.go
```diff
@@ -61,7 +63,43 @@ func (storage Storage) Put(ctx context.Context, url string, path string, class t
...
+			Url:       url,
+			Path:      path,
+			SizeBytes: fileSize(path),
+		})
+	case Directory:
+		var files []hostfile
+		files, err = walkFiles(path)
+
+		for _, f := range files {
+			u := strings.TrimSuffix(url, "/") + "/" + f.rel
```
buchanae: We lost something here. The local backend was using `path.Join`.

adamstruck: This ends up being equivalent.

buchanae: It's not equivalent for windows.

adamstruck: I don't think we have windows support anyways

buchanae: We don't need to be making it more difficult by handling file paths the wrong way.

adamstruck: technically its a file URL at this point. According to https://en.wikipedia.org/wiki/File_URI_scheme this would still work for windows

buchanae: I can live with that

PR #292, opened 2017-11-03
Updated HTTP client to more closely match GRPC client

### LINKED ISSUE #288: Client library path?
The HTTP Restful client code currently is 
```
import "github.com/ohsu-comp-bio/funnel/cmd/client"
```
The `cmd` subdirectory seems out of place, more appropriate location would be
```
import "github.com/ohsu-comp-bio/funnel/client"
```

### LINKED ISSUE #289: Client Create task takes []byte?
The client create task is declared as:
```
func (c *Client) CreateTask(msg []byte) (*tes.CreateTaskResponse, error) {
```

Why does it take a `[]byte` byte record, rather then a `tes.Task` ?

adamstruck:
closes #288 
closes #289


client/client.go
```diff
@@ -94,20 +93,26 @@ func (c *Client) ListTasks(req *tes.ListTasksRequest) (*tes.ListTasksResponse, e
...
+	m := &jsonpb.Marshaler{}
+	msg, err := m.MarshalToString(task)
 	if err != nil {
-		return nil, fmt.Errorf("Not a valid Task message: %s", err)
+		return nil, fmt.Errorf("error marshaling task message: %v", err)
 	}
 
 	// Send request
-	r := bytes.NewReader(msg)
+	r := bytes.NewReader([]byte(msg))
```
buchanae: Is is possible to marshal directly to the request, or otherwise skip the step of marshaling to string?

buchanae:
Looks good. Thanks for doing this.

client/client.go
adamstruck: The data needs to be provided via `io.Reader` per https://golang.org/pkg/net/http/#NewRequest.  I can't think of how to get around this.

buchanae: Ya, I thought request was a writer, like response is, but it's not. Still, a small simplification:
```
var b bytes.Buffer
mar.Marshal(b, &task)
```

Issue #294, opened 2017-11-03
webdash: default sort for node list


buchanae:
The node list doesn't have a default sort, which makes the list rearrange on each automatic refresh tick. This is slightly annoying, hard to read, and sometimes the links jump out from under your cursor.

adamstruck:
The node list should be sorted by state by default.

https://github.com/ohsu-comp-bio/funnel/blob/master/webdash/app.js#L77

PR #297, opened 2017-11-04
webdash: node list sorting and list size

buchanae:
Sorts the node list by ID and increases the page size in elasticsearch (which defaults to 10) so that nodes aren't missing.


buchanae:
Closes #294

buchanae:
Obsolete by #300 which remove ng-table.

PR #298, opened 2017-11-04
Fixes to concurrent node state writes

buchanae:
This contains fixes for managing node state, in particular assigned tasks. There was a bug where some assigned tasks were being set to initializing but were never started by the node/worker. This was because concurrent writes were overwriting the node record, corrupting it, and deleting the assigned task.

Using Elasticsearch's version control helped resolve this.

I have a test in mind for this, but haven't included it yet.

Also, it's based on #296, so the diff is showing changes from that branch.


server/elastic/scheduler.go
```diff
@@ -130,6 +146,8 @@ func (es *Elastic) ListNodes(ctx context.Context, req *pbs.ListNodesRequest) (*p
 	res, err := es.client.Search().
 		Index(es.nodeIndex).
 		Type("node").
+		Version(true).
+		Size(1000).
```
buchanae: Try to make unlimited

buchanae: According to this: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-from-size.html

Elasticsearch does have a max size of 10K. I could add another zero there, but returning that many records would probably hit other issues such as grpc response size limits. I think the only way forward is pagination, which I'd like to defer to another issue. We're nowhere near having 1000 nodes in production yet.

adamstruck: OK sounds fine to me

buchanae: #303

Issue #284, opened 2017-10-31
webdash: Display execution start and end time in web UI


kellrott:
Executor logs include 

https://github.com/ga4gh/task-execution-schemas/blob/ad5a5977f703c8c5825fa8ef7decec513906c6fd/task_execution.proto#L261

and

https://github.com/ga4gh/task-execution-schemas/blob/ad5a5977f703c8c5825fa8ef7decec513906c6fd/task_execution.proto#L266

but these values don't seem to be displayed in the web UI

Issue #280, opened 2017-10-25
webdash: Style and auto-deselection


kellrott:
Something about the funnel manager website causes selected text to become deselected. It makes it very difficult to copy-and-paste text from logs

buchanae:
This is probably related to the auto-refresh

Issue #255, opened 2017-10-09
webdash: Dashboard redirects to "not found" incorrectly


buchanae:
When the dashboard can't find a task during it's polling, even though it exists, the dashboard will redirect to the not found page.

PR #300, opened 2017-11-04
webdash: style, cleanup, pagination, fixes

### LINKED ISSUE #293: webdash: current task count in node list
I've found that for administration, it would be really nice if the dashboard included a little more info in the node list, such as how many tasks it currently has running. This would allow me to quickly scan to see how busy the cluster is.

buchanae:
Closes #293, #284, #280, #255. Also makes a bunch of style changes, hopefully you agree it's cleaner.

Might be a few things left to clean up, not sure at the moment.
- [x] check for unused dependencies


buchanae:
Also closes #294

buchanae:
The latest commit adds content negotiation so that `/v1/tasks` works as a dashboard URL as well as an API endpoint.

adamstruck:
I noticed that when there are multiple pages of results tasks are not returned in sorted order. They are sorted within a single page, but then the second page ends up having the newest tasks. 

Also I am getting an HTTP handler error:

```
server               HTTP handler error
error                Not Found
time                 2017-11-06T09:11:45-08:00
url                  /favicon.ico
```

adamstruck:
404 handling is broken

buchanae:
Make pagination a button

buchanae:
I think I've addressed the outstanding TODOs

adamstruck:
This will be a big improvement to the dashboard. 

Everything looks good.  Once the tests pass I'd be happy to merge this in.

PR #304, opened 2017-11-07
MongoDB

adamstruck:


worker/mongodb_taskreader.go
```diff
@@ -0,0 +1,40 @@
...
+// }
+
+// // State returns the current state of the task.
+// func (r *MongoDBTaskReader) State() (tes.State, error) {
+// 	t, err := r.client.GetTask(context.Background(), &tes.GetTaskRequest{
+// 		Id:   r.taskID,
+// 		View: tes.TaskView_MINIMAL,
+// 	})
+// 	return t.GetState(), err
+// }
```
buchanae: Maybe we want a generic task reader that just wraps a TES server or client? e.g.
```
m := NewMongoTES(conf)
reader := TaskReader(m)

r := NewRPCClient(conf)
reader := TaskReader(r)
```

server/mongodb/scheduler.go
```diff
@@ -0,0 +1,88 @@
...
+
+	return &pbs.PutNodeResponse{}, err
+}
+
+// GetNode gets a node
+func (db *MongoDB) GetNode(ctx context.Context, req *pbs.GetNodeRequest) (*pbs.Node, error) {
+	var node pbs.Node
+	err := db.nodes.Find(bson.M{"id": req.Id}).One(&node)
+	if err != nil {
+		if err == mgo.ErrNotFound {
```
buchanae: Could this error check be the same `DeleteNode` below?

server/mongodb/scheduler.go
```diff
@@ -0,0 +1,88 @@
...
+// and status updates, such as completed tasks. The server responds with updated
+// information for the node, such as canceled tasks.
+func (db *MongoDB) PutNode(ctx context.Context, node *pbs.Node) (*pbs.PutNodeResponse, error) {
+	var existing pbs.Node
+	err := db.nodes.Find(bson.M{"id": node.Id}).One(&existing)
+	if err != nil && err != mgo.ErrNotFound {
+		return nil, err
+	}
+
+	if existing.GetVersion() != 0 && node.Version != existing.GetVersion() {
```
buchanae: This is the tricky part of the scheduler. This version check alone isn't enough because there's not a transaction. You can use either optimistic concurrency control, or a document transaction here.

server/mongodb/scheduler.go
```diff
@@ -0,0 +1,88 @@
...
+
+// QueueTask adds a task to the scheduler queue.
+func (db *MongoDB) QueueTask(task *tes.Task) error {
+	return nil
+}
+
+// ReadQueue returns a slice of queued Tasks. Up to "n" tasks are returned.
+func (db *MongoDB) ReadQueue(n int) []*tes.Task {
+	var tasks []*tes.Task
+	err := db.tasks.Find(bson.M{"state": tes.State_QUEUED}).Select(basicView).All(&tasks)
```
buchanae: Can you select only `n` tasks? Also, probably want to order them by ID`

worker/mongodb_taskreader.go
adamstruck: I had to comment this out due to a circular import process. I suspect you would have the same problem in elastic.  A generic task reader seems like a good idea.

server/mongodb/task.go
```diff
@@ -0,0 +1,142 @@
...
+func (db *MongoDB) ListTasks(ctx context.Context, req *tes.ListTasksRequest) (*tes.ListTasksResponse, error) {
+	var pageSize = 256
+	if req.PageSize != 0 {
+		pageSize = int(req.GetPageSize())
+		if pageSize > 2048 {
+			pageSize = 2048
+		}
+		if pageSize < 50 {
+			pageSize = 50
+		}
```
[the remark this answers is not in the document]
adamstruck: OK will do.

server/mongodb/scheduler.go
adamstruck: Yeah this is just my first pass.  We need tests to check for the right behavior here. Right now this will pass the basic tests in `node_health_test.go`.  As far as I can tell these are the only node / scheduler related tests would be run for elasticsearch / mongodb using the current testing strategy.

server/mongodb/scheduler.go
adamstruck: Yeah I forgot about limiting to `n`.  They are ordered by ID by default.

server/mongodb/scheduler.go
adamstruck: sure

server/mongodb/events.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/golang/protobuf/ptypes"
+	"github.com/ohsu-comp-bio/funnel/events"
+	"github.com/ohsu-comp-bio/funnel/proto/tes"
+	"golang.org/x/net/context"
+	"gopkg.in/mgo.v2/bson"
+)
+
+// CreateEvent creates an event for the server to handle.
+func (db *MongoDB) CreateEvent(ctx context.Context, req *events.Event) (*events.CreateEventResponse, error) {
+	return nil, fmt.Errorf("CreateEvent - Not Implemented")
```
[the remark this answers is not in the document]
adamstruck: done

tests/e2e/server_cmd_test.go
```diff
@@ -15,7 +15,10 @@ func TestServerRunManualPanic(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	go server.Run(ctx, conf)
```
adamstruck: this was failing with:
```
fatal error: newproc: function arguments too large for new goroutine
```

buchanae: Weird, didn't know there was a goroutine limit. Would it be copying the config structure that's too large? But removing it changes the test, right?

Maybe try
```
go func() {
  server.Run(ctx, conf)
}()
```

adamstruck: Yeah thats what I changed it to.  The error channel just carried over from debugging.

adamstruck:
This PR is ready for review.

worker/worker.go
```diff
@@ -48,11 +48,9 @@ func (r *DefaultWorker) Run(pctx context.Context) {
 
 	task, run.syserr = r.TaskReader.Task()
 
-	if run.ok() {
-		r.Event.State(tes.State_INITIALIZING)
-	}
-
+	r.Event.State(tes.State_INITIALIZING)
```
buchanae: This is a surprising change to be sneaking in to a mongo database PR

buchanae:
work in progress patch for scheduler
```
index 5b346bb..e875c9f 100644
--- a/server/mongodb/scheduler.go
+++ b/server/mongodb/scheduler.go
@@ -32,22 +32,24 @@ func (db *MongoDB) ReadQueue(n int) []*tes.Task {
 // and status updates, such as completed tasks. The server responds with updated
 // information for the node, such as canceled tasks.
 func (db *MongoDB) PutNode(ctx context.Context, node *pbs.Node) (*pbs.PutNodeResponse, error) {
+ q := bson.M{"id": node.Id}
+
+ if node.GetVersion() != 0 {
+   q["version"] = node.GetVersion()
+ }
+
  var existing pbs.Node
- err := db.nodes.Find(bson.M{"id": node.Id}).One(&existing)
+ err := db.nodes.Find(q).One(&existing)
  if err != nil && err != mgo.ErrNotFound {
    return nil, err
  }

- if existing.GetVersion() != 0 && node.Version != existing.GetVersion() {
-   return nil, fmt.Errorf("Version outdated")
- }
-
  err = scheduler.UpdateNode(ctx, db, node, &existing)
  if err != nil {
    return nil, err
  }

- _, err = db.nodes.Upsert(bson.M{"id": node.Id}, node)
+ _, err = db.nodes.Upsert(q, node)

  return &pbs.PutNodeResponse{}, err
 }
```

Issue #302, opened 2017-11-06
cli/task/list: pretty print


buchanae:
We should pretty print our `funnel task list` output, because otherwise the output isn't very useful for reading:

![screen shot 2017-11-06 at 1 12 36 pm](https://user-images.githubusercontent.com/262647/32464265-52604a3a-c2f4-11e7-9b93-c879151b80e5.png)

PR #306, opened 2017-11-08
paging for 'funnel task list'

adamstruck:


buchanae:
👍 

Looks like this also closes #302

PR #318, opened 2017-11-10
server: indents for JSON responses

### LINKED ISSUE #155: `funnel task get` dumps JSON message with missing exit code log when exit code is zero.

buchanae:
I added JSON pretty printing, because why not.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -30,12 +30,8 @@ func NewClient(address string) *Client {
 		client: &http.Client{
 			Timeout: 30 * time.Second,
 		},
-		Marshaler: &jsonpb.Marshaler{
-			EnumsAsInts:  false,
-			EmitDefaults: false,
-			Indent:       "\t",
-		},
-		Password: password,
+		Marshaler: &tes.Marshaler,
+		Password:  password,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


adamstruck:
Please squash these commits.

buchanae:
Done

adamstruck:
I think that this PR breaks the expected behavior when requesting a particular task view.

$ curl -X GET http://localhost:8000/v1/tasks/b84u9fnpbjg5tks1p35g?view=MINIMAL
```
{
   "id": "b84u9fnpbjg5tks1p35g",
   "state": "QUEUED",
   "name": "",
   "description": "",
   "inputs": [
   ],
   "outputs": [
   ],
   "resources": null,
   "executors": [
   ],
   "volumes": [
   ],
   "tags": {
   },
   "logs": [
   ],
   "creationTime": ""
```

buchanae:
Ah, good point, that does seem strange, even though the data isn't actually being transmitted.

adamstruck:
The cmd/task CLI commands use their own marshaller. The return value of curl and the funnel cli should be consistent.

buchanae:
I've removed the `EmitDefaults`. I'll have to revisit that. What's left here is pretty printing (indenting) JSON by default.

buchanae:
Added marshaler to the `tes` package and shared with the rest of code for more consistency.

Closes #323

buchanae:
Unfortunately, this doesn't pretty print in the browser, but it does improve curl output. I'll have to follow up on the browser later.

Issue #325, opened 2017-11-12
autoscaling: rewrite


buchanae:
For many reasons, the current autoscaler implementations are problematic:

- they don't share code
- they're missing max/min resource limits
- they don't scale down
- the Scaler interface complicates the Scheduler interface and has caused confusion
- they make VM and Funnel node configuration difficult to maintain
- they compete with existing, robust autoscalers built into platforms such as GCE, AWS, OpenStack, etc.
- they don't include node health checking
- they really haven't been needed yet
- they're single region/zone
- they handle large batches of tasks inefficiently
- they don't allow manual scaling
- they don't handle multiple VM types
- the openstack scaler is untested and broken
- they don't allow easy reconfiguration without restarting the Funnel server
- the code has added a lot of overhead to maintaining the scheduler, tests, mocks, etc.

Most of this is my fault for not following existing autoscaler designs. I made scaling part of the per-task scheduling cycle, instead of using a simple metric such as task queue size.

I think there is still a need for an intelligent autoscaler that fits Funnel's needs:

- existing autoscalers tend to expect a homogenous workload, so scaling a wide variety of instance types is not straightforward
- existing autoscalers have a limited set of metric sources, which doesn't include Funnel task queue
- builtin autoscaling in OpenStack is not always installed/available

I think it would be very worthwhile to rewrite and simplify the autoscaler code with a simpler design:

- make a clean break from the scheduler loop
- use a metric instead of being part of the task scheduling request cycle. This makes it easy to run the autoscaler at a consistent interval.
- integrate with existing scaling resources as much as possible. e.g. use GCE's instance groups.
- make scaling up/down calls as simple as possible. This means that instance templates should probably be configured externally. This provides flexibility and easier maintenance at the cost of a more complex deployment.
- autoscaler should live as a separate service. This will make it easier to scale the Funnel servers horizontally.
- perhaps the autoscaler could be it's own project. Using a simple metric as input allows this concept to live without much knowledge of TES/Funnel at all. I think this could be a useful tool for other cases.

PR #326, opened 2017-11-12
scheduler: remove scaler, GCE, OpenStack, and manual

buchanae:
For the reasons listed in #325, I think it's best we remove the autoscaling code for now.

This PR removes over 1.5K lines of code that was mostly unused, broken, and getting in the way of features we're actively developing such as event writers, pluggable components, end-to-end encryption and security, and configuration.

This also substantially simplifies the scheduler, since it doesn't need multiple backends now.


adamstruck:
The example config and website need to be updated.

config/default-config.yaml
```diff
@@ -61,7 +60,7 @@ Server:
     # Write logs to this path. If empty, logs are written to stderr.
     OutputFile: ""
 
-# The scheduler is used for the GCE, Openstack, and Manual compute backends. 
+# The scheduler is used for the Openstack, and Manual compute backends.
```
adamstruck: There is no openstack backend anymore

config/default-config.yaml
```diff
@@ -112,7 +111,7 @@ Scheduler:
 
 
 # The name of the active compute backend
-# Available backends: local, htcondor, slurm, pbs, gridengine, manual, gce, openstack, aws-batch
+# Available backends: local, htcondor, slurm, pbs, gridengine, manual, openstack, aws-batch
```
adamstruck: There is no openstack backend anymore

adamstruck:
Did you test that the GCE deployment tools still work?

buchanae:
I also forgot to build the examples bundle. Updated.

Issue #327, opened 2017-11-12
compute: handle cancel in batch/HPC


buchanae:
@adamstruck Does AWS Batch and HPC handle canceling a queued task?

adamstruck:
No - the cancel signal is currently handled by the funnel worker.

PR #328, opened 2017-11-12
events: Task Created event and writer interface cleanup

buchanae:
Builds on #326. The diff: https://github.com/buchanae/funnel/compare/remove-scaler...buchanae:task-created-event

There's a few things going on here:
- create task event #314 
- unified event writer interface #313 
- removed the Close() method from the `events.Writer` interface.
- replaced `compute.Backend` with `events.Writer`. Removed `db.WithComputeBackend`
- abstracted `CreateTask` and `CancelTask` away from database code
- removed `scheduler.Database` in favor of `scheduler.TaskQueue` and `pbs.SchedulerService`
- remove lots of unimplemented methods from database code

The `writer.Close()` was a quick hack for Kafka that resulted in nearly all writers have an noop Close method. I clean this up here by calling `kafka.Close()` in `cmd/worker/run.Run`. This meant changing the `scheduler.WorkerFactory` interface.

The rest is about making use of the `task created` event, which simplified a lot of the connections between database, scheduler, compute backend, etc. It should also allow backends to more easily handle a cancel event for #327


buchanae:
Twice as many lines deleted as added, maybe that's a good sign :)

tests/scheduler/node_test.go
```diff
@@ -82,7 +89,7 @@ func TestManualBackend(t *testing.T) {
 	go n.Run(ctx)
 
 	tasks := []string{}
-	for i := 1; i <= 10; i++ {
+	for i := 0; i < 1; i++ {
```
adamstruck: I think more than one task should be run here

cmd/server/run.go
```diff
@@ -37,84 +39,123 @@ func Run(ctx context.Context, conf config.Config) error {
...
-		db, err = mongodb.NewMongoDB(conf.Server.Databases.MongoDB)
-	}
-	if err != nil {
-		return nil, fmt.Errorf("error occurred while connecting to or creating the database: %v", err)
-	}
+		e, err := elastic.NewElastic(conf.Server.Databases.Elastic)
+		if err != nil {
+			return nil, dberr(err)
+		}
+		if err := e.Init(ctx); err != nil {
```
adamstruck: why wasn't init absorbed into the constructor method for elastic like you did for the other databases?

server/elastic/new.go
```diff
@@ -0,0 +1,99 @@
...
+	} else if !exists {
+		if _, err := es.client.CreateIndex(name).Body(body).Do(ctx); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Init initializing the Elasticsearch indices.
+func (es *Elastic) Init(ctx context.Context) error {
```
adamstruck: This should be a private method

server/dynamodb/events.go
```diff
@@ -282,10 +276,14 @@ func (db *DynamoDB) WriteContext(ctx context.Context, e *events.Event) error {
...
-	return err
+	return checkErrNotFound(err)
 }
 
-// Close closes the writer.
-func (db *DynamoDB) Close() error {
-	return nil
+func checkErrNotFound(err error) error {
+	if e, ok := err.(awserr.RequestFailure); ok {
+		if e.StatusCode() == 400 {
```
adamstruck: This is too generic. 400 can mean lots of things. 

http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html#Programming.Errors.MessagesAndCodes

worker/interfaces.go
```diff
@@ -1,16 +1,9 @@
 package worker
 
 import (
-	"context"
 	"github.com/ohsu-comp-bio/funnel/proto/tes"
 )
 
-// Worker is a type which runs a task.
-type Worker interface {
```
adamstruck: Why was this interface deleted?

server/dynamodb/events.go
buchanae: Hrm, I guess the best I can do is check for "ValidationException" then?

worker/interfaces.go
buchanae: It's only needed by the node, so it lives there now, and replaced the WorkerFactory.

cmd/server/run.go
buchanae: I guess the other option is to pass context to NewElastic. I didn't want to hunt down other NewElastic calls at the time I wrote this, so this was the quick solution

tests/scheduler/node_test.go
buchanae: Agreed, this was for debugging.

server/dynamodb/events.go
adamstruck: I think you need to change how state change events are handled here.  Probably would be better to do something more similar to mongo or elastic here. 

I think that these updates will create an item in the database if it doesn't exist (http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html)

adamstruck: we should probably have tests to see what happens in these event handlers when that task does not exist

server/tes.go
```diff
@@ -0,0 +1,71 @@
+package server
```
adamstruck: This looks like it will be very useful.

server/dynamodb/events.go
adamstruck: I agree. It was just something I realized thinking about this code. 

The main difference here is the error handling. I think the error handling should be reverted back to how it was before. 

The state transition code need to change to be able to handle canceling an unknown task.  I think this could be achieved in nearly identical manner to how it is implemented in mongo and elastic.

buchanae: Do you mean that the state change should use `GetTask` plus `tes.ValidateTransition` in order to be more like mongo and elastic?

As I read it, the `attemptItem` and `indexItem` updates here will fail with a `ValidationError` if the task doesn't exist, which is similar to mongo's and elastic's `ErrNotFound`. Failing if the task doesn't exist and returning `tes.ErrNotFound` seems like the right behavior, no?

buchanae: Or, maybe you mean that all updates should have an added conditional check that the task exists?

adamstruck: > Do you mean that the state change should use GetTask plus tes.ValidateTransition in order to be more like mongo and elastic?

Yes.

> As I read it, the attemptItem and indexItem updates here will fail with a ValidationError if the task doesn't exist, which is similar to mongo's and elastic's ErrNotFound. Failing if the task doesn't exist and returning tes.ErrNotFound seems like the right behavior, no?

I don't think they will fail in that case. 

> Or, maybe you mean that all updates should have an added conditional check that the task exists?

Yeah all of these updates may need something along the lines of `ConditionExpression: "attribute_exists(<taskId>)` so that they will fail if the task doesn't exist.

PR #330, opened 2017-11-13
travis: try caching dependencies

buchanae:


Makefile
```diff
@@ -104,6 +106,10 @@ test-mongodb:
 
 # Run backend tests
 test-backends:
```
adamstruck: this step should depend on `cross-compile`

adamstruck:
finished up in #345

PR #345, opened 2017-11-16
Travis cache submodules and parallelize

adamstruck:


adamstruck:
This all appears to be working now. I built this multiple times in travis with manual restarts, rebases, and new commits. Seems to pass every time.

buchanae:
great work!

PR #354, opened 2017-11-18
Google Cloud Datastore

buchanae:
Based on #328 

This adds a [Google Cloud Datastore](https://cloud.google.com/datastore/docs/) backend.

From command line tests, the Task API seems to be working. I haven't run the e2e tests yet.

As always, figuring out the data model took some careful thought, but I think it's headed in the right direction. I designed the query and storage layout with the goal of adding memcache support in the near future, hence the use of KeysOnly() and GetMulti() instead of Query() and Project(). I've also tried to plan ahead for ListTasks filtering, which I hope will be easy to add.

This does not implement the scheduler API.

There's some things left to do:
- [x] general code and docs cleanup
- [x] add the emulator and e2e tests to the test suite
- [x] store input content (did I mention how annoying this field is to implement?)

In a future PR, I will:
- add support for [Memcache](https://cloud.google.com/appengine/articles/scaling/memcache)
- ~add a non-RPC event writer to the worker, which also integrates with memcache~ done
- ~implement system logs~ done

~Until #328 is merged, the better diff is here:
https://github.com/ohsu-comp-bio/funnel/compare/buchanae:task-created-event...buchanae:datastore-2~


buchanae:
Made some progress here. Stored input content, and got the tests running. Fixed some bugs. Just a little left to do.

buchanae:
Assuming the builds pass, this is ready

PR #356, opened 2017-11-21
Support for non-Amazon S3 services

### LINKED ISSUE #257: storage/swift: upload fails on large objects
Getting upload errors "Too Large Object" from swift.

Related info:
https://godoc.org/github.com/ncw/swift#Connection.DynamicLargeObjectCreate

https://github.com/ncw/rclone/commit/cc7b9af50ead0ed209afd3e70fd29975180eb8db

### LINKED ISSUE #332: storage: google storage should be on by default
This is the config currently required to enable google bucket storage:
```
Worker:
  Storage:
    GS:
      - FromEnv: true
```

We've discussed that storage should be on by default where possible, and we'll have an additional flag to disable it explicitly.

### LINKED ISSUE #338: storage/s3: support non-aws S3 and support URLs containing an endpoint
Funnel's S3 client is the AWS S3 client. We're not 100% this works outside of AWS.

There's at least one issue with using it against Cleversafe S3:
```
error                unable to find bucket bioinformatics_scratch's region not found
```

This is coded here:
https://github.com/ohsu-comp-bio/funnel/blob/master/storage/s3.go#L48

We have another request to allow the URL to potentially contain an endpoint. I guess that looks like `s3://endpoint/bucket_name/object/path`. I don't fully understand yet how Funnel can determine the difference between that and `s3://bucket_name/object/path`. Maybe we need to require one or the other?

### LINKED ISSUE #352: storage/swift: env detection causes system error
```
local                System error
attempt              0
error                failed to config Swift storage backend: Timeout when reading or writing data
index                0
taskID               b8867r2rl6qjhvin3lhg
time                 2017-11-18T08:46:46-08:00
timestamp            2017-11-18T08:46:46.16705505-08:00
```

I think what's happening is that ApplyEnvironment() is trying to connect to swift, which doesn't exist, and it returns an error which fails the task.

adamstruck:
_New features:_
  - support multiple S3 implementations (ceph, minio, etc)
  - support for S3 URIs of the form `s3://<endpoint>/<bucket>/<key>`
  - standardized cloud storage backends and e2e tests

_Fixes issues:_
- closes #257 
- closes #332 
- closes #338 
- closes #352 

_Known issues:_
  - All non amazon S3 backends are hardcoded to use the V2 signing method. 

Example config:

```YAML
Worker:
  Storage:
    AmazonS3:
      Key: ""
      Secret: ""
    GenericS3:
      - Endpoint: "localhost:8080"
        Key: "access_key"
        Secret: "secret_access_key"
```


adamstruck:
TODO:
- [ ] add new tests to travis build

adamstruck:
closes #352

storage/local.go
```diff
@@ -70,29 +62,27 @@ func (local *LocalBackend) Get(ctx context.Context, url string, hostPath string,
...
+}
 
+// Supports indicates whether this backend supports the given storage request.
+// For the LocalBackend, the url must start with "file://" be in an allowed directory
+func (local *LocalBackend) Supports(rawurl string) error {
+	path := getPath(rawurl)
+	ok := strings.HasPrefix(path, "/")
 	if !ok {
-		return fmt.Errorf("local storage does not support put on %s", url)
+		return fmt.Errorf("Must provided an absolute path")
```
buchanae: typo

buchanae: Also, please ensure the rawurl is included in the final error message, whether here or by the worker code.

storage/local.go
```diff
@@ -70,29 +62,27 @@ func (local *LocalBackend) Get(ctx context.Context, url string, hostPath string,
...
 	}
-
 	if !isAllowed(path, local.allowedDirs) {
-		return fmt.Errorf("Can't access file, path is not in allowed directories:  %s", url)
+		return fmt.Errorf("Can't access file, path is not in allowed directories:  %s", path)
```
buchanae: This should stay `url` (or `rawurl`) so that the error relates directly to the task message, without transformation.

storage/swift.go
```diff
@@ -122,49 +120,67 @@ func (sw *SwiftBackend) get(src io.Reader, hostPath string) error {
...
 	reader, oerr := os.Open(hostPath)
 	if oerr != nil {
 		return oerr
 	}
 
-	writer, err := sw.conn.ObjectCreate(url.bucket, url.path, true, "", "", nil)
+	var writer io.WriteCloser
+	var err error
+
+	// TODO turn on checkHash
```
buchanae: Yes, please do turn this on. If I understand correctly, it was previously on and was critical for preventing corrupted files in smc-het.

storage/generic_s3.go
```diff
@@ -0,0 +1,105 @@
...
+	case File:
+		err := s3.client.FGetObjectWithContext(ctx, url.bucket, url.path, hostPath, minio.GetObjectOptions{})
+		if err != nil {
+			return err
+		}
+
+	case Directory:
+		// Create a done channel.
+		doneCh := make(chan struct{})
+		defer close(doneCh)
```
buchanae: That is a strange API choice by minio...

buchanae:
A few minor issues, but otherwise tons of great work here.

Also, this: https://github.com/rabix/bunny/issues/382#issuecomment-347526813

storage/local_test.go
```diff
@@ -11,13 +11,16 @@ import (
...
+	l := LocalBackend{allowedDirs: []string{"/path"}}
+
+	err := l.Supports("file:///path/to/foo.txt")
+	if err != nil {
+		t.Fatal("Expected file:// URL to be supported", err)
 	}
 
-	if !l.Supports("/path/to/foo.txt", "/host/path", tes.FileType_FILE) {
-		t.Fatal("Expected normal file path to be supported")
+	err = l.Supports("/path/to/foo.txt")
```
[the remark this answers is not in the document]
adamstruck: I didn't change any behavior here.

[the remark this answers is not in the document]
buchanae: Is this test contrary to this comment? https://github.com/ohsu-comp-bio/funnel/pull/291#pullrequestreview-74161155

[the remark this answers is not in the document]
adamstruck: Not necessarily.  We can support both file URIs and linux style absolute paths.  Windows users are just restricted to file URIs. 

Either way, we have supported this for a long time now.  This PR isn't changing any behavior in the local backend.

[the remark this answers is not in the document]
buchanae: Gotcha, thanks. Filed #364 to clean this up.

PR #368, opened 2017-11-30
storage: http(s) backend

adamstruck:
Supports GET requests for public files.


storage/storage.go
```diff
@@ -30,7 +40,7 @@ type Backend interface {
 	PutFile(ctx context.Context, url string, path string) error
 	// Determines whether this backends supports the given request (url/path/class).
 	// A backend normally uses this to match the url prefix (e.g. "s3://")
-	Supports(url string) error
+	Supports(url string, class tes.FileType, op Operation) error
```
buchanae: This seems fine, but we could also try supporting a put.

storage/storage.go
```diff
@@ -16,6 +16,16 @@ import (
...
 )
 
+// Operation is a type used to define funnel storage operations
+type Operation int
+
+const (
+	// Get represents the operation Get(ctx context.Context, rawurl string, hostPath string, class tes.FileType) error
+	Get Operation = iota
+	// Put represents the operation Put(ctx context.Context, rawurl string, hostPath string, class tes.FileType) error
+	Put
```
buchanae: Probably better if these are private

buchanae: Oh, except, I guess the Backend interface needs to be private too, then.

storage/storage.go
adamstruck: I cant think of a use case where there is going to be a server accepting anonymous POST requests.

storage/storage.go
adamstruck: The worker accesses these in the input and output validation steps. 

The other option is to have a two methods: `SupportsGet` and `SupportsPut`.

storage/storage.go
buchanae: One case would be an http endpoint inside a private network, but it doesn't always need to be anonymous.

PR #369, opened 2017-12-01
Dynamo event cleanup

buchanae:
Follow up from the discussion here: https://github.com/ohsu-comp-bio/funnel/pull/328#pullrequestreview-77555809

Note that I duplicated some code from `GetTask`, which will be de-duplicated by #328 

Happy to add other events to the test if needed.

### DIFF
--- a/server/dynamodb/events.go
+++ b/server/dynamodb/events.go
@@ -8,6 +8,8 @@ import (
 	"github.com/ohsu-comp-bio/funnel/events"
 	"github.com/ohsu-comp-bio/funnel/proto/tes"
 	"golang.org/x/net/context"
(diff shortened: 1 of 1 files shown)


buchanae:
Requiring that an event fail to be written when the task doesn't exist actually adds some unnecessary complexity. For example, stdout/err are often stored separately from the task, and some database conditional checks don't go across row/document, so writing those events would require an extra call to GetTask to ensure it exists.

So, I guess we can do best effort for each database backend, but I don't think we need to go out of the way and hurt performance. At the very least, I guess we should require that state changes fail, since #328 uses that to implement `CancelTask`.

Issue #344, opened 2017-11-15
system_logs: implement for all databases


adamstruck:
- [x] boltdb
- [x] elastic
- [x] dynamo
- [x] mongo

PR #386, opened 2017-12-08
System log event handling

adamstruck:


buchanae:
closes #344

webdash/task.html
```diff
@@ -12,138 +12,149 @@ <h5 style="text-align: center">Not Found</h5>
   </div>
 
   <div ng-show="loaded">
-  <table class="task-desc-table">
```
buchanae: What happened here? I thought you were just adding a section?

webdash/style.scss
```diff
@@ -213,3 +208,8 @@ main h6 {
     border-color: #FF3D00;
   }
 }
+
+.btn-cancel-simple {
+  @extend .btn-cancel;
+  margin-left: 20px;
+}
```
buchanae: I kept that button small on purpose. You didn't like it?

webdash/task.html
adamstruck: The indentation was off in the file which is why the change looks so large.

webdash/style.scss
adamstruck: Actually the size was better.  My gripe with it was that it didn't match the style of the rest of the website.

buchanae:
Looks great, nice work!

PR #396, opened 2017-12-13
worker: TaskReader.Task() retries and timeout

adamstruck:


adamstruck:
It is probably better to implement retry logic at the service level. 

Looks like only the RPC client was missing this feature so it was added in #416

PR #397, opened 2017-12-14
storage: retries for swift

buchanae:


adamstruck:
Should we add a configurable timeout to both the Swift and GenericS3 configs?

adamstruck:
`TestMultiS3Storage` is timing out.  What I think is happening is the storage backend is trying to decide which S3 backend to use by looping through and calling `SupportsGet`.  The `retrier` is going to spend 15 minutes in this case retrying the `SupportsGet` call on the first S3 backend it encounters.

adamstruck:
Looking at the minio client documentation it should be retrying requests already... We may just need to adjust these variables & constants.

https://godoc.org/github.com/minio/minio-go#pkg-constants

adamstruck:
Also the swift connection class has a retries field:

https://godoc.org/github.com/ncw/swift#Connection

buchanae:
Ok, thanks for digging into that. For swift, as I understand it, those retries only apply to auth service calls, where this retrier is handling a more broad class of issues, such as object corrupted and service unavailable.

Looks like the library hard codes RequestOpts in most (all?) places, so this may be an internal class leaking out. https://github.com/ncw/swift/blob/master/swift.go#L1455

adamstruck:
For swift the retries look like they apply to all "GET" and "HEAD" calls as well as auth calls. 

https://github.com/ncw/swift/blob/067578e41c8dd06bf557d02a92a77bd35e97c05a/swift.go#L724-L738

buchanae:
Ah, good find. I guess this could be tested more then, but the experience is clear: whatever the swift library is doing isn't enough. 503 and object corrupted errors have been very common.

buchanae:
> Should we add a configurable timeout to both the Swift and GenericS3 configs?

Maybe at some point. It's not straightforward to describe what this config value would mean. I tried to add this and realized that the retrier might not make sense for larger objects, because `ExponentialBackoff.MaxElapsedTime` (15 minutes) could be shorter than one attempt to upload an object; one attempt to upload would take up the entire backoff time. Not sure how to resolve that.

This solves part of the problem. I'd prefer to get this merged and revisit these caveats later.

PR #404, opened 2017-12-16
web/dashboard: better formatting of stdout/err

buchanae:
Turns this:
screen shot 2017-12-15 at 7 40 01 pm

Into this:
screen shot 2017-12-15 at 7 37 08 pm


adamstruck:
This breaks text wrapping.  Also the same formatting rules should probably apply to stdout, stderr, and input content.

buchanae:
Added wrapping and removed the code tag.

PR #409, opened 2017-12-28
Catch and log more worker / compute backend errors

### LINKED ISSUE #25: Handle dead workers and nodes
Incomplete jobs on dead workers should probably be rescheduled. Also, should dead/gone workers be automatically cleaned up from the database at some point (after a configurable time)?

### LINKED ISSUE #28: Watch for jobs that never start

### LINKED ISSUE #118: Report when jobs cannot be scheduled on HTCondor, Slurm, etc.

### LINKED ISSUE #327: compute: handle cancel in batch/HPC

### LINKED ISSUE #429: worker should exit with a non-zero code on error

adamstruck:
The overall goal here was to improve logging when worker processes do not run correctly. Whether that be from HTCondor never starting the worker due to resource constraints, a bad job definition in AWS Batch, or a dead Funnel Node.  

Secondary to this was adding the capability to cancel tasks that are in external queues such as AWS Batch.  This feature was included here since both features required the external queue ID to be stored as metadata in Funnel. 

Originally I wanted to break this up into multiple PRs but ended up having trouble doing this since many of the code change were complimentary. 

closes #25 
closes #28 
closes #118 (minus grid engine)
closes #327 
closes #429


adamstruck:
This does not handle the case where workers loose connection to the database or server.  For that we need a task health ping or something of the sort.

adamstruck:
closing in favor of #438 and #439

PR #413, opened 2017-12-29
worker: cleanup workdir

### LINKED ISSUE #385: worker: clean up working directory
I thought we had this already, but apparently it was only in a special deployment. To avoid running out of disk space, workers should clean up their working directory when finished. This should be configurable via `Worker.LeaveWorkDir`

This should carefully consider how this interacts with local file storage

adamstruck:
closes #385


worker/worker.go
```diff
@@ -181,6 +171,11 @@ func (r *DefaultWorker) Run(pctx context.Context) {
 	if run.ok() {
 		r.Event.Outputs(outputs)
 	}
+
+	// cleanup workdir
+	if !r.Conf.LeaveWorkDir {
+		os.RemoveAll(r.Mapper.WorkDir)
+	}
```
buchanae: Should be in `defer`

buchanae: Might even want it as a method on `FileMapper`?

worker/worker.go
```diff
@@ -89,15 +88,6 @@ func (r *DefaultWorker) Run(pctx context.Context) {
...
 
-	// Create working dir
-	var dir string
-	if run.ok() {
-		dir, run.syserr = filepath.Abs(filepath.Join(r.Conf.WorkDir, task.Id))
-	}
-	if run.ok() {
-		run.syserr = fsutil.EnsureDir(dir)
-	}
-
```
buchanae: Would be better in `FileMapper.MapTask`

worker/file_mapper.go
```diff
@@ -40,15 +40,22 @@ type Volume struct {
...
-	return &FileMapper{
+func NewFileMapper(dir string) (*FileMapper, error) {
+	workdir, err := filepath.Abs(dir)
+	if err != nil {
+		return nil, err
+	}
+	err = fsutil.EnsureDir(workdir)
+	if err != nil {
+		return nil, err
+	}
```
buchanae: Would be better in `MapTask`

config/config.go
```diff
@@ -113,6 +113,9 @@ type Worker struct {
 	UpdateRate time.Duration
 	// Max bytes to store in-memory between updates
 	BufferSize int64
+	// Normally the worker cleans up its working directory after executing.
+	// This option disables that behavior.
+	LeaveWorkDir bool
```
buchanae: Need to update default-config.yml

PR #420, opened 2018-01-06
storage/local: add error messages. create|wronly copy mode

buchanae:
While attempting to run [bunny](https://github.com/rabix/bunny) using a FUSE + Swift setup, I ran into a failure in the Funnel local storage backend. The FUSE library only supports create, read-only, and write-only file modes, while Go's `os.Create` uses a read-write mode.

This PR includes all the error messages I needed to add to track down the problem, and a change to use create + write-only for local file copy.

This PR needs a little work maybe, putting it here so I don't forget about it.

### DIFF
--- a/storage/local.go
+++ b/storage/local.go
@@ -113,11 +113,11 @@ func copyFile(source string, dest string) (err error) {
 	// Open source file for copying
 	sf, err := os.Open(source)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to open source file for copying: %s", err)
 	}
-	df, err := os.Create(dest)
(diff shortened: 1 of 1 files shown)


storage/local.go
```diff
@@ -141,44 +141,45 @@ func linkFile(source string, dest string) error {
...
-		return same, err
+		return false, fmt.Errorf("failed to stat src file: %s", err)
 	}
 	dfi, err := os.Stat(dest)
 	if os.IsNotExist(err) {
-		return same, nil
+		return false, nil
 	} else if err != nil {
-		return same, err
+		return false, fmt.Errorf("failed to state dest file: %s", err)
```
adamstruck: typo: state -> stat

PR #439, opened 2018-01-22
compute: cancel queued tasks; log failed Compute.Submit calls

### LINKED ISSUE #28: Watch for jobs that never start

### LINKED ISSUE #118: Report when jobs cannot be scheduled on HTCondor, Slurm, etc.

### LINKED ISSUE #327: compute: handle cancel in batch/HPC

adamstruck:
Cancel queued tasks in AWS Batch, HTCondor, Slurm, etc.

closes #327

### DIFF
--- a/compute/batch/backend.go
+++ b/compute/batch/backend.go
@@ -14,23 +14,29 @@ import (
 )
 
 // NewBackend returns a new local Backend instance.
-func NewBackend(batchConf config.AWSBatch) (*Backend, error) {
+func NewBackend(batchConf config.AWSBatch, reader tes.ReadOnlyServer, writer events.Writer) (*Backend, error) {
 
 	sess, err := util.NewAWSSession(batchConf.AWSConfig)
 	if err != nil {
 		return nil, fmt.Errorf("error occurred creating batch client: %v", err)
 	}
 
-	return &Backend{
-		client: batch.New(sess),
-		conf:   batchConf,
-	}, nil
+	b := &Backend{
+		client:   batch.New(sess),
+		conf:     batchConf,
+		event:    writer,
+		database: reader,
+	}
+
+	return b, nil
 }
 
 // Backend represents the local backend.
 type Backend struct {
-	client *batch.Batch
-	conf   config.AWSBatch
+	client   *batch.Batch
+	conf     config.AWSBatch
+	event    events.Writer
+	database tes.ReadOnlyServer
 }
 
 // WriteEvent writes an event to the compute backend.
@@ -38,13 +44,18 @@ type Backend struct {
 func (b *Backend) WriteEvent(ctx context.Context, ev *events.Event) error {
 	switch ev.Type {
 	case events.Type_TASK_CREATED:
-		return b.Submit(ev.GetTask())
+		return b.Submit(ctx, ev.GetTask())
+
+	case events.Type_TASK_STATE:
+		if ev.GetState() == tes.State_CANCELED {
+			return b.Cancel(ctx, ev.Id)
+		}
 	}
 	return nil
 }
 
 // Submit submits a task to the AWS batch service.
-func (b *Backend) Submit(task *tes.Task) error {
+func (b *Backend) Submit(ctx context.Context, task *tes.Task) error {
 	req := &batch.SubmitJobInput{
 		// JobDefinition: aws.String(b.jobDef),
 		JobDefinition: aws.String(b.conf.JobDefinition),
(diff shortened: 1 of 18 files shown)


adamstruck:
This PR also covers creating system log events when the Compute.Submit method fails.

compute/batch/backend.go
```diff
@@ -68,10 +79,52 @@ func (b *Backend) Submit(task *tes.Task) error {
...
+
+	return b.event.WriteEvent(
+		ctx, events.NewMetadata(task.Id, 0, map[string]string{"awsbatch_id": *resp.JobId}),
+	)
+}
+
+// Cancel removes tasks from the AWS batch job queue.
+func (b *Backend) Cancel(ctx context.Context, taskID string) error {
+	var task *tes.Task
+	var err error
```
buchanae: nit: these seem unnecessary. the `x := Y()` pattern is the best option.

compute/batch/backend.go
```diff
@@ -68,10 +79,52 @@ func (b *Backend) Submit(task *tes.Task) error {
...
+
+	task, err = b.database.GetTask(
+		ctx, &tes.GetTaskRequest{Id: taskID, View: tes.TaskView_FULL},
```
buchanae: Probably want the basic view.

compute/batch/backend.go
```diff
@@ -68,10 +79,52 @@ func (b *Backend) Submit(task *tes.Task) error {
...
+		return err
+	}
+
+	// only cancel tasks in a QUEUED state
+	state := task.GetState()
+	if state != tes.State_QUEUED {
+		return nil
+	}
+
+	backendID := getAWSTaskID(task)
```
buchanae: There's a small chance that the batch ID isn't there. Probably want to handle that case.

compute/hpc_backend.go
```diff
@@ -14,45 +15,95 @@ import (
...
+	)
+}
+
+// Cancel cancels a task via "qdel", "condor_rm", "scancel", etc.
+func (b *HPCBackend) Cancel(ctx context.Context, taskID string) error {
+	var task *tes.Task
+	var err error
+
+	task, err = b.Database.GetTask(
+		ctx, &tes.GetTaskRequest{Id: taskID, View: tes.TaskView_FULL},
```
buchanae: Basic view here too?

compute/hpc_backend.go
```diff
@@ -14,45 +15,95 @@ import (
...
+	return b.Event.WriteEvent(
+		ctx,
+		events.NewMetadata(task.Id, 0, map[string]string{fmt.Sprintf("%s_id", b.Name): backendID}),
+	)
+}
+
+// Cancel cancels a task via "qdel", "condor_rm", "scancel", etc.
+func (b *HPCBackend) Cancel(ctx context.Context, taskID string) error {
+	var task *tes.Task
+	var err error
```
buchanae: same pattern as above

compute/hpc_backend.go
```diff
@@ -14,45 +15,95 @@ import (
...
+
+	task, err = b.Database.GetTask(
+		ctx, &tes.GetTaskRequest{Id: taskID, View: tes.TaskView_FULL},
+	)
+	if err != nil {
+		return err
+	}
+
+	// only cancel tasks in a QUEUED state
+	state := task.GetState()
```
buchanae: GetState isn't necessary, since State is an int value. `if task.State != tes.Queued` would do.

compute/hpc_backend.go
```diff
@@ -14,45 +15,95 @@ import (
...
+	if state != tes.State_QUEUED {
+		return nil
+	}
+
+	backendID := getBackendTaskID(task, b.Name)
+	if backendID != "" {
```
buchanae: Looks like you handle the missing ID here, just need to make Batch match.

compute/hpc_backend.go
```diff
@@ -14,45 +15,95 @@ import (
...
+		cmd := exec.Command(b.CancelCmd, backendID)
+		return cmd.Run()
+	}
+
+	return fmt.Errorf("failed to get %s_id for task %s", b.Name, taskID)
```
buchanae: nit: the pattern I like is that exceptional conditions, such as an empty ID, are handled in the body of the if-statement, and the regular flow stays at the root level. The code here is the opposite.

PR #440, opened 2018-01-23
compute: reconcile task states with external queues

### LINKED ISSUE #28: Watch for jobs that never start

### LINKED ISSUE #118: Report when jobs cannot be scheduled on HTCondor, Slurm, etc.

adamstruck:
based off of #439 

https://github.com/adamstruck/funnel/compare/compute-cancel...adamstruck:compute-reconcile

closes #28 
closes #118 (minus grid engine)


compute/htcondor/backend.go
```diff
@@ -32,3 +44,96 @@ func extractID(in string) string {
...
+				output = append(output, &compute.HPCTaskState{ID: id, TESState: tes.SystemError, State: state, Reason: "Funnel worker exited with non-zero status"})
+			}
+		}
+	}
+	return output, nil
+}
+
+func checkIdleStatus(id string) error {
+	cmd := exec.Command("condor_q", "-analyze", id)
+	stdout, _ := cmd.Output()
```
buchanae: Might as well return the error. Who knows, we might need it later.

compute/pbs/backend.go
```diff
@@ -27,3 +38,78 @@ func NewBackend(conf config.Config, reader tes.ReadOnlyServer, writer events.Wri
...
+
+	idSet := make(map[string]interface{})
+	for _, i := range ids {
+		idSet[i] = nil
+	}
+
+	cmd := exec.Command("qstat", "-x")
+	stdout, err := cmd.Output()
+	if err != nil {
+		return nil, err
```
buchanae: can you wrap these errors? I've noticed with other pieces of code with unwrapped errors that I struggle to hunt down the source line in funnel

adamstruck:
Since this PR increases the likelihood of concurrent state updates I recommend merging #482 first.

tests/slurm.config.yml
```diff
@@ -1 +1,4 @@
 Compute: "slurm"
+Slurm:
+  DisableReconciler: false
```
buchanae: Sorry to be picky, but these should match the actual default config, i.e. they should be true.

PR #453, opened 2018-01-24
api/tes: ListTasks state filter

buchanae:
Taking a swing at ListTasks filtering. Intended to be quick and dirty to get this ball rolling, will refine towards the end of this PR.


adamstruck:
Nice work.  I have two comments:

1) Can we change `state_filter` in the `ListTasksRequest` to `state`
2) Add a CLI flag to `funnel task list`

buchanae:
Ok, address the feedback and added a filter to the webdash UI. The UI style is pretty ugly, I'll revisit that in a future PR.

cmd/task/list.go
```diff
@@ -10,10 +10,9 @@ import (
 )
 
 // List runs the "task list" CLI command, which connects to the server,
-// calls ListTasks() and requests the given task view. Results may be filtered
-// client-side using the "taskState" and "taskName" args. Output is written
-// to the given writer.
-func List(server, taskView, pageToken string, pageSize uint32, all bool, writer io.Writer) error {
+// calls ListTasks() and requests the given task view.
+// Output is written to the given writer.
+func List(server, taskView, pageToken, stateFilter string, pageSize uint32, all bool, writer io.Writer) error {
```
adamstruck: `stateFilter` is never used.

adamstruck:
For the webdash, it would be nice if the url was updated as filters are applied.

Makefile
```diff
@@ -95,59 +94,59 @@ start-elasticsearch:
...
+	@go test -timeout 120s ./tests/gridengine -funnel-config `pwd`/tests/gridengine.config.yml
 
 test-pbs-torque:
 	@docker pull ohsucompbio/pbs-torque
-	@go test -timeout 120s ./tests/pbs -funnel-config $(CONFIGDIR)/pbs.config.yml
+	@go test -timeout 120s ./tests/pbs -funnel-config `pwd`/tests/pbs.config.yml
 
 test-amazon-s3:
-	@go test ./tests/storage -funnel-config $(CONFIGDIR)/s3.config.yml -run TestAmazonS3
+	@go test ./tests/storage -funnel-config `pwd`/s3.config.yml -run TestAmazonS3
```
adamstruck: Here and several places below: 
```
`pwd`/ -> `pwd`/tests
```

buchanae:
> For the webdash, it would be nice if the url was updated as filters are applied.

I agree. Just for the sake of time, I'm going to punt on this for now though.

https://github.com/ohsu-comp-bio/funnel/issues/473

cmd/task/list.go
```diff
@@ -26,12 +25,20 @@ func List(server, taskView, pageToken string, pageSize uint32, all bool, writer
 
 	output := &tes.ListTasksResponse{}
 
+	state := tes.Unknown
+	if stateInt, ok := tes.State_value[stateFilter]; ok {
```
adamstruck: The logic here should probably be a closer match to the `getTaskView` utility above so we get case insensitivity and an informative error message.

buchanae:
This is ready to go. I've restarted the TravisCI build multiple times.

PR #454, opened 2018-01-25
logger/grpc: set most grpc logs to debug

buchanae:
The grpc logs are a pain. In the screenshot below you can see they overwhelm the TravisCI test logs with pages of junk. This is an attempt to clean those up by moving most things to the debug level.

Perhaps one things that's missing is to call SetGRPCLogger in all the top-level commands such as server, node, and worker.

screen shot 2018-01-24 at 4 53 51 pm

### DIFF
--- a/cmd/node/node.go
+++ b/cmd/node/node.go
@@ -60,6 +60,7 @@ func newCommandHooks() (*cobra.Command, *hooks) {
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			log := logger.NewLogger("node", conf.Logger)
+			logger.SetGRPCLogger(log)
 			ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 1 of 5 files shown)


adamstruck:
I am confused as to why those Info level grpc logs are showing up. As of #437 only warnings are logged.

adamstruck:
I think we should call `SetGRPCLogger` from the top level commands rather than using the `init` function. That way we can turn on the Debug level logs.

buchanae:
> I am confused as to why those Info level grpc logs are showing up. As of #437 only warnings are logged.

me too. I guess it's these lines:
https://github.com/ohsu-comp-bio/funnel/pull/437/files#diff-c0793adc313c89b9e72678e66dc1edc3L29

> I think we should call SetGRPCLogger from the top level commands rather than using the init function. That way we can turn on the Debug level logs.

I agree. I'll add that.

PR #465, opened 2018-01-27
events/pubsub: google pub/sub event writer

buchanae:


buchanae:
#453 adds a docker container for Google emulators, which is much easier/cleaner. When that's merged, I'll fix this up.

config/config.go
```diff
@@ -166,6 +167,12 @@ type Kafka struct {
 	Topic   string
 }
 
+// PubSub configures access to Google Cloud Pub/Sub for task event reading/writing.
+type PubSub struct {
+	Topic   string
+	Project string
```
adamstruck: This should also have `CredentialsFile string` to be consistent with the other google components.

buchanae: Fixed in https://github.com/ohsu-comp-bio/funnel/pull/465/commits/8a7743b41a4f0e9bffb88fcb6be190bbb651efc5

adamstruck: It doesn't look like your `NewPubSubWriter` and `ReadPubSub` methods make use of the new field.

PR #466, opened 2018-01-27
tes: remove executors

buchanae:
Here's a quick and dirty pass at removing multiple executors, just to see what it might look like.

Improvements I noticed:

- less code, less to maintain
- less dealing with nested logs.logs arrays. much easier to flatten all that.
- we'll probably never get around to making `attempt` work anyway
- simpler events and logs
- `funnel run` can be simplified, removing a lot of complex code. Might be able to merge this back into a cobra command
- do we really need stdout/err paths in the task? can't the command handle that?
- do we really need volumes? can't the container handle that?
- the event generators were unnecessary and a lot of boilerplate
- worker is substantially simpler. stepWorker went away.
- this would probably make elasticsearch/kibana/other searches easier, since a lot of nesting is removed. nesting is usually where databases start to have trouble.
- we say this is "easy to implement, just a for loop", but in hindsight I'd say it turned out to take quite a bit of code (although it's fairly simple code).
- the concept of a task is definitely easier to explain without executors.


proto/tes/validate.go
```diff
@@ -24,35 +24,21 @@ func (v ValidationError) Error() string {
...
+  }
 
-		if exec.Stdin != "" && !strings.HasPrefix(exec.Stdin, "/") {
-			errs.add("Task.Executors[%d].Stdin: must be an absolute path", i)
-		}
+  if len(t.Command) == 0 {
+    errs.add("Task.Command: required, but empty")
+  }
 
-		if exec.Stdout != "" && !strings.HasPrefix(exec.Stdout, "/") {
```
adamstruck: stdout validation is still needed

proto/tes/validate.go
```diff
@@ -24,35 +24,21 @@ func (v ValidationError) Error() string {
...
-			errs.add("Task.Executors[%d].Stdout: must be an absolute path", i)
-		}
+  if t.Workdir != "" && !strings.HasPrefix(t.Workdir, "/") {
+    errs.add("Task.Workdir: must be an absolute path")
+  }
 
-		if exec.Stderr != "" && !strings.HasPrefix(exec.Stderr, "/") {
```
adamstruck: stderr validation is still needed

events/events.proto
```diff
@@ -21,18 +21,16 @@ message SystemLog {
...
+  START_TIME = 2; 
+  END_TIME = 3;
+  OUTPUTS = 4;
+  METADATA = 5;
+  EXIT_CODE = 8;
+  STDOUT = 11;
+  STDERR = 12;
   SYSTEM_LOG = 13;
-  TASK_CREATED = 14;
+  CREATED = 14;
```
adamstruck: CREATED -> CREATION_TIME

proto/tes/tes.pb.go
```diff
@@ -11,10 +11,8 @@ It has these top-level messages:
 	Task
 	Input
 	Output
-	Executor
```
adamstruck: you didnt check in the proto file

worker/worker.go
```diff
@@ -241,36 +239,19 @@ func (r *DefaultWorker) fixLinks(mapper *FileMapper, basepath string) {
 }
 
 // openLogs opens/creates the logs files for a step and updates those fields.
-func (r *DefaultWorker) openStepLogs(mapper *FileMapper, s *stepWorker, d *tes.Executor) error {
+func (r *DefaultWorker) openStepLogs(ctx context.Context, event *events.TaskWriter, mapper *FileMapper, cmd *DockerCommand, t *tes.Task) error {
```
adamstruck: rename openStepLogs -> openLogs

worker/worker.go
```diff
@@ -241,36 +239,19 @@ func (r *DefaultWorker) fixLinks(mapper *FileMapper, basepath string) {
...
-	if d.Stdout != "" {
-		s.Command.Stdout, err = mapper.CreateHostFile(d.Stdout)
-		if err != nil {
-			s.Event.Error("Couldn't prepare log files", err)
-			return err
-		}
-	}
-
-	// Create file for task stderr
-	if d.Stderr != "" {
```
adamstruck: I think this is still needed

worker/worker.go
```diff
@@ -241,36 +239,19 @@ func (r *DefaultWorker) fixLinks(mapper *FileMapper, basepath string) {
...
-	if d.Stdin != "" {
-		s.Command.Stdin, err = mapper.OpenHostFile(d.Stdin)
-		if err != nil {
-			s.Event.Error("Couldn't prepare log files", err)
-			return err
-		}
-	}
-
-	// Create file for task stdout
-	if d.Stdout != "" {
```
adamstruck: I think this is still needed

adamstruck:
It is nice how much simpler things got with these changes.

erikvdbergh:
Hi Alex,

I would +1 this PR, simplifying TES down to the core. I think the main reason to have multiple executors was for caching reasons, i.e. pulling down data only once for related tasks. Do you have ideas about an alternative strategy for this?

We've been toying with the concept of 'linked tasks', where you somehow indicate to the TES API that separate tasks are linked so the data can stick around. Not sure how this would be implemented.

Love to hear your thoughts.

buchanae:
> Do you have ideas about an alternative strategy for this?

Yes, ideas, and a few experiments, but nothing concrete yet. In my mind, linking tasks is essentially form of workflow description, and a good workflow engine should be the component responsible for scheduling workflow steps efficiently.

A bare bone, TES-based workflow description would be an adjacency list of tasks:
```
{
  "tasks": {
    "task-1": { "TES task here" },
    "task-2": { "TES task here" },
  },
  "dependencies": {
    "task-1": ["task-2", "task-3", "etc"],
  },
}
```

Efficiency and caching, especially with respect to file caching, requires a service to track which files are on which nodes. Another element is how a scheduler schedules tasks to a node, balancing the tradeoff between how long the task will wait vs how much time/money/effort is saved by waiting vs redownloading the file.

There are other use cases too, which might have different solutions. Generally, I'd say they're all complex enough that they deserve attention from separate services.

PR #471, opened 2018-01-30
api/tes: ListTask tag filters

adamstruck:
Based on #453 

https://github.com/buchanae/funnel/compare/list-filtering...adamstruck:tag-filter?expand=1


adamstruck:
The webdash still needs to be updated to allow tag filtering

client/client.go
```diff
@@ -85,6 +85,12 @@ func (c *Client) ListTasks(ctx context.Context, req *tes.ListTasksRequest) (*tes
 		addString(v, "state", req.State.String())
 	}
 
+	if req.GetTags() != nil {
```
buchanae: You don't need this. It's ok to range over a nil map: https://play.golang.org/p/e-ynTY8LdVx

cmd/task/list.go
```diff
@@ -30,12 +31,22 @@ func List(server, taskView, pageToken, stateFilter string, pageSize uint32, all
 		return err
 	}
 
+	tags := make(map[string]string)
+	for _, v := range tagsFilter {
+		parts := strings.Split(v, "=")
```
buchanae: This doesn't handle the case where a key or value contains an "=" sign. Do we care about that case?

server/boltdb/task.go
```diff
@@ -161,10 +161,19 @@ func (taskBolt *BoltDB) ListTasks(ctx context.Context, req *tes.ListTasksRequest
...
 
+	taskLoop:
 		for ; k != nil && i < pageSize; k, _ = c.Prev() {
 			task, _ := getTaskView(tx, string(k), req.View)
 			if req.State != tes.Unknown && req.State != task.State {
-				continue
+				continue taskLoop
+			}
+
+			if req.Tags != nil {
```
buchanae: Also not needed.

server/server.go
```diff
@@ -40,11 +40,11 @@ func newDebugInterceptor(log *logger.Logger) grpc.UnaryServerInterceptor {
...
-		log.Debug(
-			"responding: "+info.FullMethod,
-			"resp", resp,
-			"err", err,
-		)
+		// log.Debug(
+		// 	"responding: "+info.FullMethod,
+		// 	"resp", resp,
+		// 	"err", err,
+		// )
```
buchanae: oops?

cmd/task/list.go
adamstruck: do we handle that case in `funnel run`? I took this pattern from there.

adamstruck:
so apparently dynamo doesn't allow the following:
```
tes.Task{
    Tags: map[string]string{"foo": ""},
}
```

In the above case the `Tags` field is dropped during serialization.

adamstruck:
Ok this PR should be ready to go.

buchanae:
How many of the databases will work with tasks created before this PR?

Also, does Dynamo drop tags with an empty value still?

adamstruck:
I was wrong about Dynamo which is why I deleted the comment. All of the databases except datastore should work with tasks created before this PR.

PR #480, opened 2018-02-02
cmd/run: remove default workdir

### LINKED ISSUE #287: cmd/run: overrides workdir by default
I was surprised to find that `funnel run` results in a task that overrides workdir to `/opt/funnel`. (yes, I'm surprised even though I likely wrote this code.)

Should we consider removing that?

buchanae:
closes #287

### DIFF
--- a/cmd/run/flags.go
+++ b/cmd/run/flags.go
@@ -129,10 +129,6 @@ func newFlags(v *flagVals) *pflag.FlagSet {
 
 // Set default flagVals
 func defaultVals(vals *flagVals) {
-	if vals.workdir == "" {
-		vals.workdir = "/opt/funnel"
-	}
-
 	if vals.container == "" {
 		vals.container = "alpine"
 	}


adamstruck:
For some reason I thought we did this on purpose to allow relative paths or something?

buchanae:
I think we should merge this. Here's what I tested:

`funnel run 'echo hi > out'`

Doesn't work because the working directory is a read-only filesystem (in the default alpine image)

`funnel run 'echo hi > $out' -o out=output.txt`

Works.

`funnel run 'md5sum $in > $out' -i in=input.txt -o out=output.txt`

Works.

Having `funnel run` make it impossible to use the default workdir set by an image seems like the worst case. If there's some other critical reason why we need `funnel run` to force the workdir that we're forgetting, we'll have to change it back later.

adamstruck:
Logic seems sound to me.

Issue #475, opened 2018-01-31
storage/swift: tweak retry config


buchanae:
We're getting frequent failures from our swift cluster (503 unavailable) and the retrier isn't doing enough. In it's current default configuration, it will only retry for a couple seconds.

Just how long to retry is an open question, but I'd say we should err on the side of retrying for too long, rather than too briefly. These are large files being uploaded, created as the result of processes that potentially take multiple days, so it's worth retrying for hours even.

More technically, we could drop the `MaxRetries` in favor of `MaxElapsedTime`, which I think is a better configuration. It's more difficult to calculate the max amount of time you're retrying from a count number with exponential backoff involved.

buchanae:
Somewhat related: it'd be nice to know exactly how the underlying ncw/swift library handles 503 errors. Does one 503 error mean the entire file needs to be retried? Can it have better handling of segment uploads to the whole upload progress isn't lost?

PR #482, opened 2018-02-06
upgraded the databases to handle concurrent task state updates

### LINKED ISSUE #476: databases: concurrent writes can cause state transition issues
Essentially in the EventWriter implementations for State events we call GetTask and use the reported state to validate the transition. When multiple State events happen concurrently, GetTask may return the same State in each instance and proceed with the update to the database.   

To solve this we need to make the state transitions more transactional. 

Notes:
* Datastore uses transactions already so we just need to shift some code around. 
* Dynamo transaction library - https://github.com/fumin/dtx

adamstruck:
closes #476 

I opted to only enforce version checks for the state events for now. If we need to expanded it to all event types the pieces are now there to make it easy. 

I ended up abstracting the `retrier` from the storage package into a general utility.


adamstruck:
@buchanae since I generalized the retrier in this PR should I also address #475 since that code is changing here anyways?

PR #484, opened 2018-02-07
api/tes: WIP example of custom resource types

buchanae:
Work in progress example of a possible solution to https://github.com/ga4gh/task-execution-schemas/issues/84


buchanae:
replaced by #537

PR #485, opened 2018-02-08
Kid's First Dev

adamstruck:
A PR to track changes I make as I get a dev version of funnel up and running for the project. 

ChangeLog:
* api/cancel: compute event needs to be handled first since it checks the current state prior to issuing the backend specific request
* server logging on compute.Cancel error was messed up
* Event writer constructors (e.g `NewDynamoDB`) should not try to create tables, etc in the worker process.


adamstruck:
Split into #486 and #487

PR #506, opened 2018-03-14
storage: use an io.Reader that utilizes context

### LINKED ISSUE #400: storage swift / local / gs: cancelling context does not stop download/upload

adamstruck:
closes #400


buchanae:
wow. Did you copy that code from something reliable? We have a lot of storage bugs as it is, so I'm concerned about adding this much complexity.

adamstruck:
Just about all of it is an exact copy of `io.Copy` as seen here: https://golang.org/src/io/io.go

buchanae:
Instead of forking io.Copy, could we close the io.Reader when the context is canceled?

buchanae:
Or create a wrapper that implements io.Reader, as done by:
- https://gist.github.com/dchapes/6c992bf3e943934462509338cd213e99
- https://github.com/jbenet/go-context/blob/master/io/ctxio.go

adamstruck:
Generally it isn't considered goo practice to store context on an object like they are doing in those examples. That being said, both of those examples could also work. Then again, I don't see anything wrong with what I implemented. It certainly seems to work fine.

buchanae:
It's fine to bend the rule about storing context in this case, and the resulting code will be much more simple, and more reusable too since it's not Copy-specific.

PR #507, opened 2018-03-14
auth: update default config and website docs

buchanae:
Missed some docs and default config for #505


config/default.go
```diff
@@ -29,6 +29,7 @@ func DefaultConfig() Config {
 		DisableHTTPCache:    true,
 		RPCClientTimeout:    time.Second * 60,
 		RPCClientMaxRetries: 10,
+		User:                "funnel",
```
adamstruck: I don't think there should be a default value for user.

adamstruck:
Also you should update the default-config.yaml file.

PR #508, opened 2018-03-15
fix for stdout/err log tail collection

### LINKED ISSUE #430: Truncated stderr output
The stderr field is emitting this:

```
]['DateCompleted']
KeyError: 'DateCompleted'
```

But many other similar failing jobs have the full error output. (example: https://github.com/biostream/pubmed-transform/issues/3)

adamstruck:
closes #430

### DIFF
--- a/events/executor.go
+++ b/events/executor.go
@@ -190,7 +190,7 @@ func StreamLogTail(ctx context.Context, taskID string, attempt, index uint32, si
 
 	flush := func(buf *ring.Buffer, t Type, timeout <-chan time.Time) {
 		// Only flush if new bytes have been written to the buffer.
-		if buf.NewBytes() == 0 {
+		if buf.NewBytesWritten() == 0 {
 			return
 		}
 
@@ -210,7 +210,7 @@ func StreamLogTail(ctx context.Context, taskID string, attempt, index uint32, si
 		select {
 		case eventch <- e:
 			// The writer routine accepted the event, so reset the buffer byte count.
-			buf.ResetNewBytes()
+			buf.ResetNewBytesWritten()
 		case <-timeout:
 			// The writer was busy, do nothing.
 		}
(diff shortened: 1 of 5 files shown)


buchanae:
I ran the ring_test.go against master and they all passed. Is the real fix here the task builder?

adamstruck:
No the fix is in the ring / executor event builder.  I had to add the newBytes field and use that rather than totalWritten.  Resetting the totalWritten field was the root cause of the issue. 

Try running the new e2e test on master and you will see the issue

buchanae:
Can you add a test for the ring buffer that fails on master then?

adamstruck:
The e2e test covers the bug.  I don’t think we can write a test for just the ring buffer that could be run on both branches since I had to add a new field and method  on the structure and delete a method.

buchanae:
I don't get it. If the ring buffer is wrong somehow, a test should be able to demonstrate how it's wrong.

adamstruck:
The ResetTotalWritten method was the problem and it was removed.

buchanae:
What was wrong with ResetTotalWritten? If the tests show that the buffer always returns the string you'd expect, what is wrong with it?

adamstruck:
Run the E2E test on master and you will see the behavior it caused.  It was a method that was not originally from circbuf.   The totalWritten field is used in the Bytes() method so reseting it without reseting the cursor as well causes the wrong data to be returned. 

Removing this method prevents this bug from being possible.  This change is most easily demonstrated by the E2E test I wrote.

buchanae:
This is what I meant: https://github.com/ohsu-comp-bio/funnel/pull/509/files#diff-e3740da1387227aacaa00d8d5927b21fR198

adamstruck:
added the test, but looking at the code we probably didnt need to fork https://github.com/armon/circbuf

I think that I will remove the forked code and tests altogether.

PR #509, opened 2018-03-15
util/ring: TotalWritten shouldn't affect buffer content returned

buchanae:


adamstruck:
closing since its a duplicate of #508

PR #512, opened 2018-03-15
worker/docker: log container metadata

### LINKED ISSUE #418: worker: log docker image hash
This is more reliable that the docker image name / tag pair.

buchanae:
closes #418


buchanae:
Unfortunately, we don't have executor metadata, so logging to system logs is the easiest option for now. I suggest we take this first step, and improve that part later.

PR #515, opened 2018-03-21
Moving code

### LINKED ISSUE #333: tes clients: merge "client" package with "proto/tes" move "proto/tes" to "tes"
While writing docs for the website, I noticed that it's not easy to refer to our TES clients and utils in a single place. I think these clients are a valuable part of Funnel that can be easily reusable with a little organization.

I suggest we merge the http client into the "tes" package, and move "proto/tes" to the root of the project. The "proto" name seems unnecessary at this point.

adamstruck:
* merged "client" package with "proto/tes" move "proto/tes" to "tes"; closes #333 
* moved "proto/scheduler" files into "compute/scheduler"
* moved database code from "server" package into "database" package


compute/scheduler/node.go
```diff
@@ -6,15 +6,14 @@ import (
...
-	pbs "github.com/ohsu-comp-bio/funnel/proto/scheduler"
 	"github.com/ohsu-comp-bio/funnel/util"
 	"github.com/ohsu-comp-bio/funnel/util/fsutil"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
 
-// NewNode returns a new Node instance
-func NewNode(ctx context.Context, conf config.Config, factory Worker, log *logger.Logger) (*Node, error) {
+// NewNodeInstance returns a new Node instance
```
buchanae: I don't like this. If the proto is getting in the way, it should be in a subpackage.

adamstruck: I think we should either rename this type or the proto message type rather than use the same name for two different things.

buchanae: Ya, that is a bit confusing. How about calling the proto message something like `NodeInfo`?

adamstruck: I am good with that.

buchanae: Sweet. Otherwise this all looks great. Thanks for doing this, it's a nice cleanup.

PR #525, opened 2018-04-05
docker: allow docker base command args to be configured

buchanae:
There are cases where the standard "docker" command is not the appropriate/best command to execute for docker containers. Examples include custom docker wrappers such as exadocker, udocker, etc.

This changes allows config such as:
```
Docker:
  BaseArgs: ["sudo", "/path/to/docker-wrapper"]
```

This is what I'm thinking for allowing the docker command to be configured/overridden.

Note that since I'm touching the docker config, I modified the `RemoveContainer` flag, which was previously hard-coded to `true` and never exposed via config. This change exposes that as `config.Docker.LeaveContainer`.

Another minor tweak is that all os/exec commands in the docker executor now use CommandContext.

To do:
- [ ] update config example
- [ ] website docs
- [ ] tests

### DIFF
--- a/cmd/worker/run.go
+++ b/cmd/worker/run.go
@@ -106,6 +106,7 @@ func NewWorker(ctx context.Context, conf config.Config, log *logger.Logger) (*wo
 		Store:       store,
 		TaskReader:  reader,
 		EventWriter: writer,
+		DockerConf:  conf.Docker,
 	}
 
 	return w, nil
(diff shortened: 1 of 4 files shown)


buchanae:
I'm not planning to finish this any time soon, and I'm not sure it's the right approach anyway. In the meantime, a simple wrapper script and PATH config would do what I needed.

Issue #478, opened 2018-02-01
compute/node: duplicate task ID


buchanae:
Ran into a case in the wild where somehow a node had a duplicate task ID:
```
curl http://funnel.compbio.ohsu.edu/v1/nodes/funnel-node-3
{
  "id": "funnel-node-3",
  "resources": {
    "cpus": 12,
    "ramGb": 63.332102144,
    "diskGb": 2495.074340864
  },
  "available": {
    "cpus": 1,
    "ramGb": 63.332102144,
    "diskGb": 2495.074340864
  },
  "state": "ALIVE",
  "version": "176071",
  "taskIds": [
    "b9nrllnntf6mbcjc5pug",
    "b9nrllnntf6mbcjc5pug",
    "b9pmg3vntf6mbcjc5r50",
    "b9pmm4fntf6mbcjc5r90",
    "b9pmmpnntf6mbcjc5ra0",
    "b9pmmpvntf6mbcjc5rag",
    "b9pmmpnntf6mbcjc5r9g",
    "b9pmmpvntf6mbcjc5rbg",
    "b9pmmpvntf6mbcjc5rb0",
    "b9pmmpvntf6mbcjc5rb0",
    "b9pn91nntf6mbcjc5sl0",
    "b9pn91nntf6mbcjc5sm0",
    "b9pn91nntf6mbcjc5slg",
    "b9pn91nntf6mbcjc5sm0"
  ],
  "lastPing": "1517517332038755287"
}
```

PR #527, opened 2018-04-06
compute/node: add zone config

buchanae:
This change allows the "zone" resource to be configured for nodes, which allows a cluster of nodes to be split up into groups, which may be requested by tasks. This is useful as a crude way of partitioning a cluster, and/or for designating some nodes which have different config, such as a "debug" zone which keeps more logs/working directories.

I had some hacky version of this deployed previously, and it was very useful, but I've lost it. 

To do:
- [x] tests


buchanae:
The remaining problem here is that Elasticsearch is a terrible database for something which requires consistency, such as the scheduler. Duplicate tasks are getting assigned to nodes (#478) which is breaking the test.

I think I'm going to recommend we remove the scheduler backend from the Elasticsearch code.

PR #533, opened 2018-04-12
cmd/task/create: read tasks from stdin

buchanae:
I have a script that generates many tasks. This feature allows me to pipe those tasks directly to funnel with `python generate-tasks.py | funnel task create -`. 

In the past I've written the extra code to write these tasks to a bunch of files in a directory, which is an annoying extra step when you're doing this frequently.

To do:
- [x] cli and website docs
- [x] tests


buchanae:
Latest commit drops the hyphen usage and will read from both stdin and the file list, e.g.
`python generate-tasks.py | funnel task create`

Also, it now writes out the task IDs as they're created, instead of in a single batch at the end, which feels more responsive.

PR #534, opened 2018-04-13
tes: force minimum 1 cpu

### LINKED ISSUE #55: Default resource requirements
A Funnel worker needs some minimum amount of resources to run. 

If a task message is submitted that doesn't include resources Funnel should pick some reasonable defaults.

buchanae:
Closes #55 

This does give up the ability to request zero CPUs, but that's harder to achieve and this is a glaring issue that can cause servers to crash.

To do
- [ ] test


buchanae:
We agreed this isn't the right implementation. This should happen per compute backend.

Issue #518, opened 2018-03-22
compute/scheduler: add DrainNode request


adamstruck:
Add a mechanism that allows an administrator to request that a node stops accepting new tasks so that it can be shut down, upgraded, etc.

buchanae:
Closed by #554

Issue #517, opened 2018-03-21
compute/scheduler: scheduler does not correctly process queue


buchanae:
The scheduler only ever processes the top N queued tasks. It's possible that these tasks are not able to be scheduled, but tasks farther down in the queue may be scheduled.

For example, say there are two zones: zone 1 supports 10 tasks, zone 2 supports 5 tasks. The queue contains 100 tasks requesting zone 1, and 10 tasks requesting zone 2, in that order. The scheduler will fill up zone 1, but it will not reach the tasks requesting zone 2.

This happens here: https://github.com/ohsu-comp-bio/funnel/blob/master/compute/scheduler/scheduler.go#L96

Issue #432, opened 2018-01-18
compute/node: it is possible to have two active nodes with the same id


adamstruck:
Seems like this should be avoided as it would break the scheduler and cause tasks to run multiple times.

Is there any downside to generating a UUID on node startup?  We could add a name field to replace the functionality of specifying the node id.

Issue #299, opened 2017-11-04
scheduler: redesign node state data structure


buchanae:
Adding the Elasticsearch backend highlighted some difficulties in dealing with node state. In particular, there are two, distributed entities writing to the same record concurrently: the scheduler writes to the Node record to assign a new task, and the node writes to its Node record when it finishes a task or to update other metadata such as resources.

I've employed a few tricks to deal with distributed concurrency:
1. I introduced a `Version` field, which is used for optimistic writes.
2. The node must retry its writes in case of a version conflict.
3. In elasticsearch, I use its internal version control, because external version control comes with extra caveats that were causing problems.

Getting consistent state in elasticsearch took some work. There were subtle details in how to use versions: passing them between the server and node, unmarshaling node records and then set the version, asking for the primary shard, refreshing the index after writes, retrying requests, etc.

On one hand, I **think** I've figured out all the tricks needed and it's working. On the other, it's tricky to get right. I'd like to add a wealth of backends, and possibly this will be a hurdle for each one.

I think there's potentially a better design that uses an event log. The "assign task" and "task complete" events are independent, and they shouldn't need to worry about conflicting writes. It would be up to the reader to resolve the current state of assigned tasks based on the event log. Writes would always succeed without conflict, and would be stored in the database independently (instead of in a nested array) in order to simplify this.

I haven't fully worked out whether that design works, but wanted to get these thoughts written down while I'm currently engrossed in it. @kellrott @adamstruck thoughts?

buchanae:
After more thought, I'm fairly certain the best path forward is to have the nodes schedule their own tasks. That removes much of the complexity of having a central scheduler, and removes the concurrent updates from two, distributed processes.

PR #536, opened 2018-04-25
compute/manual: rewrite

buchanae:
Here's an experiment in redesigning the scheduler and node. The intention here is to solve many of the common issues we've had with the scheduler:

- correct task queue processing (#517)
- node management (#518, #432, #325, #180)
- node state conflicts (#478, #299, #161)

As of now, this experiment doesn't solve all of those, but it's headed in the right direction, I think.

This removes the node database in favor of a more online approach. GRPC bidirectional streaming is used to establish an online connection between the node and server. Node records are kept in memory. 

Benefits:
- The scheduler now follows the same pattern as the other compute backends, reacting to a CreateTask (aka Submit, aka TaskCreated event) call.
- Task can be immediately scheduled, which compares to modern job schedulers like Slurm in speed.
- The node state is not stored in a database, meaning less complex management of conflicts and concurrency.
- Cleaned up compute backend using interface, so that it's not event based.

There's plenty left to do here:

- store tasks in a durable (persistent) queue
- manage that queue correctly (#517)
- reimplement handling of tasks assigned to failed nodes
- handle cancelation and context correctly
- add a Drain command (#518)


buchanae:
- Added a DRAIN_NODE state, which can be sent by the server. Still needs to be hooked up to more useful admin commands.
- Added Badger as the persistent task queue.
- Added back cleaning up of GONE nodes/tasks.
- Removed the SchedulerChunk limit, instead choosing to schedule all queued tasks on each iteration, which should fix (#517) and also allow more interesting priority and named queues in the future.
- Remove ScheduleRate config. If needed, scheduling rate should be controlled automatically with some intelligent backoff. For most cases, I think scheduling very frequently should be fine, since the database is now much more lightweight (and dedicated only to the scheduler).

To do:
- having lots of issues with Ctrl+c
- add useful utils/cmd for draining node
- consider periodically (and on startup) calling ListTasks{State == QUEUED} and reconciling this with the badger queue. This might help catch edge cases, and also allow the scheduler to be move to a different server by rebuilding the queue database.

PR #551, opened 2018-05-19
cmd/storage: add get and put

buchanae:
One thing to consider: the interface for put is `put [url] [path]`, which feels backwards, but matches the `Storage` interface. Should we reverse it for the CLI? Reverse it in both the interface and CLI? Leave it?

### DIFF
--- a/cmd/storage/storage.go
+++ b/cmd/storage/storage.go
@@ -150,6 +150,62 @@ func NewCommand() *cobra.Command {
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


adamstruck:
I think you should reverse it in the CLI. I don't think it matters all that much for the interface.

buchanae:
Swapped it for the CLI. Put is now `funnel storage put [path] [url]`

PR #554, opened 2018-05-21
compute/manual: node draining

buchanae:
This PR adds a DRAIN state for nodes under the manual compute backend.

Nodes are drained by sending the linux SIGHUP signal to the node process. This can be done manually via `kill -SIGHUP <node PID>`, but in reality this is sent by `systemctl reload funnel`.


buchanae:
Please take a look.

adamstruck:
Looks good to me.

PR #560, opened 2018-05-23
docs: example systemd service file for funnel node

buchanae:
I have service files like these scattered all over the place. Committing them here so they're somewhere useful.


adamstruck:
I agree that this is useful. Can you add an example systemd file for the server as well?

Issue #470, opened 2018-01-30
api: add /metrics endpoint for prometheus metrics


buchanae:
We could get a lot of cool graphs out of prometheus if we set up some simple metrics in the prometheus format:

- counts for each task state
- counts for nodes, available resources, etc

More advanced/complex, but still very useful:

- error counts, e.g. service unavailable counts
- download/upload rates (bytes/sec?)
- server request latency

Some of these are things we already want in the dashboard.

Go library: https://godoc.org/github.com/prometheus/client_golang/prometheus

Helpful links:
https://www.robustperception.io/how-does-a-prometheus-counter-work/
https://medium.com/@griggheo/initial-experiences-with-the-prometheus-monitoring-system-167054ac439c

buchanae:
Added the basics. Will open new issues as more is desired.

PR #571, opened 2018-05-26
database/mongo: add task state counts to service info

buchanae:
Adds task state counts to the service info endpoint for mongo.


buchanae:
Added node metrics to prometheus. Closes #470

adamstruck:
A few comments:
- You need to run `dep ensure` to update Funnel's dependencies
- It would be nice if the task state counts had a place in the web dashboard
- Some docs around the prometheus metrics would be helpful

buchanae:
> Also why did you copy http.proto and annotations.proto rather than use the vendored version?

I think these aren't vendored since the switch to `dep`. I thought it was simpler to just copy these two, rather than submodule the whole repo.

buchanae:
Also needs:
- [ ] tests

buchanae:
> It would be nice if the task state counts had a place in the web dashboard

At first I added this to the task list page, but then I realized it might be confusing in conjunction with the task filters, so I put it on the service info page as a simple fix. We can revisit something more advanced in the future.

adamstruck:
looks good 👍

PR #574, opened 2018-05-30
build: use external release notes builder

### LINKED ISSUE #520: build/release: CLI version of release notes builder
I've extracted the release notes builder into a separate CLI tool and added some flags.

https://github.com/buchanae/github-release-notes

In particular, we'll need to change the `--stop-at` flag (which PR to stop reading at) per release.

buchanae:
closes #520


adamstruck:
How do you feel about making the value of the `-stop-at` flag a Makefile variable as you are doing with the version in #575?

buchanae:
> How do you feel about making the value of the -stop-at flag a Makefile variable as you are doing with the version in #575?

Sure, that's a good idea.

Issue #384, opened 2017-12-08
web/dashboard: optimize task list auto-refresh


buchanae:
The auto refresh on the task list is a excessive. Looks like it refreshes every 2 seconds, and in my dashboard that's download over 500KB per request.

I do like the idea of having the dashboard update without needing to hit refresh, but there's a lot of room for optimization here:

- increase the interval to at least 10 seconds
- use the [page visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) to only update when the tab is active (i.e. don't bother refreshing when the user isn't looking)
- refresh tasks individually, and cache tasks in a final state
- expose a task events API via HTTP and subscribe to that, pushing polling/streaming down a layer in the stack
- after the initial load, only state changes, so refreshes can use the minimal task view

https://github.com/ohsu-comp-bio/funnel/blob/master/webdash/app.js#L51

<img width="1389" alt="screen shot 2017-12-08 at 9 59 00 am" src="https://user-images.githubusercontent.com/262647/33778659-7f697c40-dbfe-11e7-8a6a-9a3897653d14.png">

buchanae:
The main ideas left here are:

- allow GetTask and ListTasks to request a specific subset of fields in the response to enable the client to get exactly what it wants without extra weight
- provide an event system which client can easily subscribe to
- server side rendering of initial page load
- configurable and pause-able refresh in the dashboard

I'm not sure any of those are worth the amount of work and complexity. The initial problem was improved a lot. 

The configurable and pause-able refresh loop is a good idea, but let's address that in a separate issue later.

PR #576, opened 2018-05-30
webdash: optimize task list

### LINKED ISSUE #547: web/dashboard: initial page load requests task list 3 times
I noticed the dashboard task list is slow to load. Digging into the network profiling, I see that the initial load does 3 simultaneous task list requests.

screen shot 2018-05-09 at 11 21 40 am

buchanae:
closes #547


buchanae:
Also, closes #384 ...maybe

I added code to cache tasks in the browser, use the minimal view where possible, and stop the refresh loop when the page/tab is in the background.

It's still not ideal, but it's a substantial improvement. Previously, our task list was requesting about 500-800MB every two seconds, even when you weren't looking. Now, it's about 500 bytes when nothing has changed and nothing is running.

The page load is still slow on the first load (although faster than before). This might be improved in a number of ways: partial list rendering, server-side rendering, etc.

Our task views are kinda clunky. The basic view can still be pretty heavy, and the list doesn't need most of that weight (~500-700KB). We could improve this substantially by having server-side support for requesting specific fields, and/or being able to subscribe to an event stream.

adamstruck:
This looks like a great improvement.  Could you apply the stoping of the background refresh loop while the page is in the background behavior to the task, node list and node pages as well?

buchanae:
> Could you apply the stoping of the background refresh loop while the page is in the background behavior to the task, node list and node pages as well?

Done

PR #578, opened 2018-05-31
tes: add ETag field to OutputFileLog

adamstruck:
@buchanae what do you think about adding this to make it easy to parse out of the task message?


buchanae:
I think this is great.

Small nit, I'd say `e_tag` should be `etag`.

It makes me wonder though, should we be more careful about continuing to add more non-standard fields to the TES endpoint?

adamstruck:
I was initially adding this to enable gaia to include the etag in its provenance graph.

However, since there are no guarantees around how the etag is calculated (or if it is even included) I am less confident in how useful this is.

PR #580, opened 2018-06-01
dynamodb: fix potential panic in tableIsAlive method

adamstruck:


database/dynamodb/util.go
```diff
@@ -196,11 +196,13 @@ func (db *DynamoDB) tableIsAlive(ctx context.Context, name string) error {
 		case <-ctx.Done():
 			return ctx.Err()
 		case <-ticker:
-			r, _ := db.client.DescribeTable(&dynamodb.DescribeTableInput{
+			r, err := db.client.DescribeTable(&dynamodb.DescribeTableInput{
 				TableName: aws.String(name),
 			})
-			if *r.Table.TableStatus == "ACTIVE" {
-				return nil
+			if err == nil {
```
buchanae: Should we log this error? If it's a permissions error or something, this might wait silently forever.

adamstruck: True... maybe it should check for a specific error type instead

buchanae:
Looks fine to me.

PR #582, opened 2018-06-02
build/release: try s3 deploy from travis ci

buchanae:
I want to see what it takes to deploy snapshots to s3 from travis


buchanae:
forget it, too much trouble

PR #593, opened 2018-06-28
storage/ftp: add FTP support

buchanae:
There's a little bit left to do here:
- auth credentials should have a place in the config file. currently the code only supports pulling user/pass from the URL.

That could be done in a future issue though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -88,6 +88,12 @@ jobs:
         - make test-pubsub
       env:
         - n=pubsub
+    - script:
+        - make start-ftp
+        - sleep 10
+        - make test-ftp
+      env:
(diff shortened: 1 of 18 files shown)


buchanae:
A quick update: this works and passes the tests in multiple environments I've tried, but only fails in Travis. So...yay.

buchanae:
The TravisCI test is fixed. The test FTP server needed help matching up the user/group ID with the host.

There are a couple items left:
- [ ] add better config so that user/password aren't in the storage URL.
- [ ] I think there's an intermittent failure where two tasks create the same directory concurrently.

buchanae:
Ready to go

PR #595, opened 2018-06-29
compute/pipelines

buchanae:
Rough draft

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,7 +2,6 @@
 
 
 [[projects]]
-  digest = "1:cd815ef8b39ff3bd0018be220392e5e08443bd38a3b432ef48a322614f305f4b"
   name = "cloud.google.com/go"
   packages = [
     "compute/metadata",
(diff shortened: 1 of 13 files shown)


buchanae:
Much of this works:
- submits pipelines to the API, which run but fail
- passes task to the pipelines as a base64 string
- cancel is implemented
- reconciler is on
- events are pulled down into task logs, various metadata can be captured
- etc

The tasks fail because the worker can't execute the containers. There's no way to mount the docker socket on the host, and there's no `--privileged` flag do docker-in-docker doesn't work. 

The only option I can think of so far is to use a sequence of `genomics.Actions` (their tes.Executors) to execute the various parts of the task (download, execute, upload, capture metadata, etc), which *should* work, but will take some effort.

Honestly though... we could just as easily make API calls directly to Compute Engine and create a Container-optimized VM per task, which is essentially what Pipelines is doing. That would be easy and we could use the Funnel worker as-is. I don't see anything in Pipelines we'd be missing with that approach.

buchanae:
Also on the TODO list:
- [ ] send config to the worker so it knows how to talk to servers, event writers, databases, etc.
- [ ] use task ID instead of base64, and/or intelligently switch between the two.

PR #615, opened 2019-06-24
Adding GO module support

kellrott:


adamstruck:
switched to modules in #626

PR #632, opened 2020-03-16
storage: add s3 presigned url support to https backend

### LINKED ISSUE #631: Issues accessing presigned s3 URLs
I can create Funnel tasks reading/writing to s3 without issues.

If I change a source URL from `s3://[...]` to a presigned URL `https://{...}.s3.amazonaws.com[..]` the task will fail. I see that in the task log the URL is listed correctly, but in the debug execution log is is changed. For example, `[...]5DT&Signature[...]` gets changed to `[...]5DT\u0026Signature[...]`, which causes the failure.

Is there anything I need to add to allow for URLs with special characters?

adamstruck:
Decouples Get / Stat methods so signed URLs can be supported. 

closes #631


adamstruck:
Note: support is for ~inputs only~

PR #633, opened 2020-03-19
Adding new docker build

kellrott:
Changes the docker build to be a staged build that is then copied into a bare alpine image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,3 +13,5 @@ VOLUME /opt/funnel/funnel-work-dir
 EXPOSE 8000 9090
 ENV PATH="/app:${PATH}"
 COPY --from=build-env  /go/src/github.com/ohsu-comp-bio/funnel/funnel /app/
+
+ENTRYPOINT ["/app/funnel"]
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+ADD ./ /go/src/github.com/ohsu-comp-bio/funnel
+RUN cd /go/src/github.com/ohsu-comp-bio/funnel && go build ./
+
+# final stage
+FROM alpine
+WORKDIR /opt/funnel
+VOLUME /opt/funnel/funnel-work-dir
+EXPOSE 8000 9090
+ENV PATH="/app:${PATH}"
+COPY --from=build-env  /go/src/github.com/ohsu-comp-bio/funnel/funnel /app/
```
adamstruck: Should we keep funnel as the entry point like in the original image?

adamstruck:
Please update the Makefile (https://github.com/ohsu-comp-bio/funnel/blob/master/Makefile#L192)

PR #635, opened 2020-03-20
add .dockerignore

adamstruck:


kellrott:
:+1:

kellrott:
Actually, is make installed in alpine?

adamstruck:
> Actually, is make installed in alpine?

It gets installed in the build-env

kellrott:
:+1:

PR #647, opened 2020-11-17
FIX: Open File leak

kellrott:
Under some deployments (local backend, but possibly others) GRPC event writers weren't closed, and so the system leaked open file handles, until it ran out under the `ulimit -n` constraint. This PR adds `Close` methods to several interfaces so that network connections and open files are closed.


kellrott:
I've checked and the pbs unit test is also happening to master. Something has changed to the docker build within the past 8 months and now the script submitted to the pbs queue in the docker container doesn't has permission to use docker (but if you exec into the container it can). This comes back as an execution error in the unit test.
I'd prefer to fix it in an other PR, rather then delay this one while trying to figure out a fix.

cmd/server/run.go
```diff
@@ -47,7 +47,7 @@ type Server struct {
 // Database represents the base funnel database interface
 type Database interface {
 	tes.ReadOnlyServer
-	events.Writer
+	//events.Writer
```
bwalsh, line 50: Why commented?  Remove, or add comment

compute/hpc_backend.go
```diff
@@ -55,6 +55,10 @@ func (b *HPCBackend) WriteEvent(ctx context.Context, ev *events.Event) error {
 	return nil
 }
 
+func (b *HPCBackend) Close() {
+	//TODO: Close things here?
```
bwalsh, line 59: If really a TODO, add test ...  If a NOOP, add comment to say so

database/dynamodb/new.go
```diff
@@ -47,3 +47,7 @@ func NewDynamoDB(conf config.DynamoDB) (*DynamoDB, error) {
 func (db *DynamoDB) Init() error {
 	return db.createTables()
 }
+
+func (db *DynamoDB) Close() {
+	//db.client.Close()
```
bwalsh, line 52: Why commented?

bwalsh (changes requested):
Minor comments

cmd/server/run.go
kellrott: Removed.

compute/hpc_backend.go
kellrott: Removed comment

database/dynamodb/new.go
kellrott: Was there from dev, when assumed there was a dynamo close method. There doesn't appear to be one, so the comment has been removed.

PR #699, opened 2023-05-12
Feat/v1.1 compliance

lbeckman314:
# Tracking TES v1.1 compliance:
 
- Compliance status — https://ohsu-comp-bio.github.io/funnel-compliance/
- Compliance repo — https://github.com/ohsu-comp-bio/funnel-compliance
  - [funnel-compliance actions](https://github.com/ohsu-comp-bio/funnel-compliance/actions)
  - [Funnel Actions](https://github.com/ohsu-comp-bio/funnel/actions)
- Compliance issues — https://github.com/ohsu-comp-bio/funnel/issues/679

## References
- [TES v1.1 Release Page](https://github.com/ga4gh/task-execution-schemas/releases/tag/v1.1)
- [ELIXR Cloud & AAI tes-compliance-suite](https://github.com/elixir-cloud-aai/tes-compliance-suite/pull/29)

PR #709, opened 2023-06-17
Add support for propagating the TES task tags as Docker labels

robertbio:
This minor change adds the values of the tags sent via the TES API when creating the task to the docker command that runs the task in the form of [Docker labels](https://docs.docker.com/engine/reference/commandline/run/#label), similar to the env values.

Benefits:
 - More efficient monitoring and management of tasks in the backend, especially in AWS Batch where different tasks can run on the same instance.


kellrott:
We're currently working on a PR to upgrade to TES 1.1 support in #699. We'll probably deal with this PR after that work is done

PR #716, opened 2023-07-06
Fix/lint

lbeckman314:


worker/worker.go
```diff
@@ -194,7 +252,7 @@ func (r *DefaultWorker) openStepLogs(mapper *FileMapper, s *stepWorker, d *tes.E
 	if d.Stdin != "" {
 		s.Command.Stdin, err = mapper.OpenHostFile(d.Stdin)
 		if err != nil {
-			s.Event.Error("Couldn't prepare log files", err)
+			err := s.Event.Error("Couldn't prepare log files", err)
```
adamstruck, line 255: This change causes the wrong error to be returned. 

In general, it might be worth considering an approach where you just log all event writer errors rather than handle them throughout the codebase. In this PR, the new error handling seems inconsistent - errors are logged, returned, ignored or even involved in a panic.

compute/local/backend.go
```diff
@@ -46,9 +67,18 @@ func (b *Backend) Submit(task *tes.Task) error {
 		return err
 	}
 
-	go func() {
-		w.Run(ctx)
+	go func() { 
+		// WaitGroup to block closing the worker until the last event is written from events/executor.go:StreamLogTail()
+		var wg sync.WaitGroup
```
adamstruck, line 72: Seems like it would make more sense to create the `WaitGroup` in worker.go and just ensure that `Run` blocks until all logs are written.

adamstruck (commented):
Nice work - I didn't look at all the changes in depth, but left a couple of general comments.

worker/worker.go
adamstruck: IIRC (and it has been a looong time) we intentionally ignored these sorts of errors. Why? I have no idea.

adamstruck: I think it was because we generally used the `MultiWriter` with the `Logger` writer ensuring that the events were recorded somewhere.

compute/local/backend.go
adamstruck: Actually, the simplest place to get the same behavior would be adding the wait group to step.go and then adding the wait call somewhere around [here](https://github.com/ohsu-comp-bio/funnel/blob/master/worker/step.go#L54-L68).

worker/worker.go
adamstruck: The Writer is actually already wrapped with one that logs all errors that occur:

https://github.com/ohsu-comp-bio/funnel/blob/226edc9f5c8e4e85b7c3c7c39f41c5d32e886c3f/cmd/worker/run.go#L58

lbeckman314: Ah, that makes sense! Changing `err :=` to `_ =` to satisfy the errcheck linter:

```diff
- err := s.Event.Error("Couldn't prepare log files", err)
+ _ = s.Event.Error("Couldn't prepare log files", err)
```

Original linting error:
```sh
➜  funnel git:(fix/lint) ✗ golangci-lint run ./... -D unused -e "deprecated"
worker/worker.go:255:17: Error return value of `s.Event.Error` is not checked (errcheck)
                        s.Event.Error("Couldn't prepare log files", err)
```

cmd/run/flags.go
```diff
@@ -207,6 +210,9 @@ func buildExecs(flags *pflag.FlagSet, vals *flagVals, args []string) {
 		}
 		return nil
 	})
+	if err != nil {
```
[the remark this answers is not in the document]
adamstruck: Similar to my previous comments you might check if these writers are wrapped with the ErrorLogger.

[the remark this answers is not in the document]
adamstruck: Maybe it would be better to update the ErrorLogger so that it doesn’t return errors so you don’t need to make these sort of changes at all.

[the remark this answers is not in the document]
adamstruck: One other option I'll throw out - add a specific exclusion for `Writer.Write` calls for the errcheck linter (see: https://golangci-lint.run/usage/false-positives/#specific-linter-excludes).

compute/scheduler/scheduler.go
```diff
@@ -63,11 +63,17 @@ func (s *Scheduler) CheckNodes() error {
 
 		if node.State == NodeState_GONE {
 			for _, tid := range node.TaskIds {
-				s.Event.WriteEvent(ctx, events.NewState(tid, tes.State_SYSTEM_ERROR))
-				s.Event.WriteEvent(ctx, events.NewSystemLog(tid, 0, 0, "info",
+				err = s.Event.WriteEvent(ctx, events.NewState(tid, tes.State_SYSTEM_ERROR))
+				if err != nil {
+					return err
```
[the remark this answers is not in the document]
adamstruck: Similar to my previous comments you might check if these writers are wrapped with the ErrorLogger. 

It might be a better approach to update the ErrorLogger so that it doesn’t return errors so you don’t need to make these sort of changes at all (you'd have to create another interface though and update usages accordingly).

One other option I'll throw out - add a specific exclusion for Writer.Write calls for the errcheck linter (see: https://golangci-lint.run/usage/false-positives/#specific-linter-excludes).

compute/local/backend.go
lbeckman314: Good call — moved the WaitGroup closer to the output event writer routine that was causing the panic after the session closed* ([executor.go](https://github.com/ohsu-comp-bio/funnel/blob/fix/lint/events/executor.go#L238-L251))

Tried adding the wait call to [step.go](https://github.com/ohsu-comp-bio/funnel/blob/master/worker/step.go#L54-L68) but encountering the same panic result — does `<-done` wait for the final event writer to complete?

---

*Session close panic that prompted adding the WaitGroup: [Actions Run](https://github.com/ohsu-comp-bio/funnel/actions/runs/5514073875/job/14928495674#step:4:438)

cmd/node/run.go
```diff
@@ -34,8 +37,8 @@ func Run(ctx context.Context, conf config.Config, log *logger.Logger) error {
 		return err
 	}
 
-	runctx, cancel := context.WithCancel(context.Background())
-	runctx = util.SignalContext(ctx, time.Nanosecond, syscall.SIGINT, syscall.SIGTERM)
+	_, cancel := context.WithCancel(context.Background())
+	runctx := util.SignalContext(ctx, time.Nanosecond, syscall.SIGINT, syscall.SIGTERM)
```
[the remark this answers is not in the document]
lbeckman314: Ah, I believe this was due to a detected linting error, specifically "ineffectual assignment to runctx":

```sh
→ golangci-lint run funnel/cmd/node/run.go
cmd/node/run.go:40:2: ineffectual assignment to runctx (ineffassign)
        runctx, cancel := context.WithCancel(context.Background())
```

Should runctx being assigned to in both cases (`context.WithCancel` and `util.SignalContext`)?

cmd/run/flags.go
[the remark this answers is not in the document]
lbeckman314: I'll change the new `err` assigns to `_` in this PR, thanks for the heads up!

compute/scheduler/scheduler.go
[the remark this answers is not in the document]
lbeckman314: Taking a look at updating the `events.ErrLogger` (or possibly just adding an exclusion to the linter). Tending toward adding the exclusion, reverting the `err` changes, and creating an issue to track adding a new interface.

compute/hpc_backend.go
```diff
@@ -73,6 +76,7 @@ func (b *HPCBackend) Submit(task *tes.Task) error {
 	cmd.Stderr = &stderr
 
 	err = cmd.Run()
+	fmt.Println("DEBUG err:", err)
```
adamstruck, line 79:
```suggestion
```

lbeckman314:
Reverted the many error checks that were triggering the `errcheck` linting error. @adamstruck you made great points and those deserve their own separate PR/issues for discussion. This better focuses the PR on the TESv1.1 compliance, will take a look at your PR's and look at integrating our work with the tes-compliance-suite (and defer to your branches if any merge conflicts arise):
- https://github.com/ohsu-comp-bio/funnel/pull/718
- https://github.com/ohsu-comp-bio/funnel/pull/587
- https://github.com/ohsu-comp-bio/funnel/pull/586

@kellrott I'll ping here when those integrations are done and ready for review!

database/boltdb/scheduler.go
```diff
@@ -8,35 +8,34 @@ import (
...
 )
 
 // queueTask adds a task to the scheduler queue.
 func (taskBolt *BoltDB) queueTask(task *tes.Task) error {
 	taskID := task.Id
 	idBytes := []byte(taskID)
 
-	err := taskBolt.db.Update(func(tx *bolt.Tx) error {
-		tx.Bucket(TasksQueued).Put(idBytes, []byte{})
+	taskBolt.db.Update(func(tx *bolt.Tx) error {
```
adamstruck, line 20: This error check shouldn’t have been removed.

database/boltdb/scheduler.go
```diff
@@ -8,35 +8,34 @@ import (
...
+		err := tx.Bucket(TasksQueued).Put(idBytes, []byte{})
```
adamstruck, line 21: Simpler if you just return this directly.

compute/local/backend.go
adamstruck: The <-done is only going to wait for the command to exit. You need to wait group call after that for it to block.

compute/scheduler/scheduler.go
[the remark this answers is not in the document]
adamstruck: Sorry I left so many unhandled errors! I wrote this back when I was still learning go, so hindsight is 20/20 I guess 😄

cmd/node/run.go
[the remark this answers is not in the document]
adamstruck: I'm thinking we meant for it to be something like:

```
subctx, cancel := context.WithCancel(ctx)
runctx = util.SignalContext(subctx, time.Nanosecond, syscall.SIGINT, syscall.SIGTERM)
```

cmd/task/get.go
```diff
@@ -19,23 +19,20 @@ func Get(server string, ids []string, taskView string, w io.Writer) error {
 
 	res := []string{}
 
-	view, err := getTaskView(taskView)
+	_, err = getTaskView(taskView)
```
adamstruck, line 22: Is this conversion from string to enum no longer needed?

database/dynamodb/util.go
```diff
@@ -363,13 +364,13 @@ func (db *DynamoDB) deleteTask(ctx context.Context, id string) error {
...
 				}
-				// TODO handle error
-				db.client.DeleteItem(item)
-			}
-			if page.LastEvaluatedKey == nil {
-				return false
+				// TODO handle error without panic
+				_, err := db.client.DeleteItem(item)
+				if err != nil {
+					log.Fatalf("failed to delete content item: %v", err)
```
adamstruck, line 370: Better than a panic:

```suggestion
					log.Printf("ERROR: failed to delete content item: %v\n", err)
```

database/mongodb/new.go
```diff
@@ -5,11 +5,13 @@ import (
 	"time"
 
 	"github.com/globalsign/mgo"
+	"github.com/ohsu-comp-bio/funnel/compute/scheduler"
 	"github.com/ohsu-comp-bio/funnel/config"
 )
 
 // MongoDB provides an MongoDB database server backend.
 type MongoDB struct {
+	scheduler.UnimplementedSchedulerServiceServer
```
adamstruck, line 14: Why is this needed?

events/computer.go
```diff
@@ -0,0 +1,36 @@
+package events
```
adamstruck, line 1: Seems like it might make more sense for this to live in the compute package?

server/tes.go
```diff
@@ -79,33 +85,61 @@ func (ts *TaskService) CancelTask(ctx context.Context, req *tes.CancelTaskReques
...
+		Id:               "org.ga4gh.funnel",
+		Name:             ts.Name,
+		Organization: map[string]string{
+			"name": "OHSU Computational Biology",
+			"url":  "https://github.com/ohsu-comp-bio",
+		},
+		Storage: []string{
+			"file:///path/to/local/funnel-storage",
+			"s3://ohsu-compbio-funnel/storage",
+		},
```
adamstruck, line 112: Seems like this would depend on the config. Google storage and ftp are also supported IIRC.

server/tes.go
```diff
@@ -79,33 +85,61 @@ func (ts *TaskService) CancelTask(ctx context.Context, req *tes.CancelTaskReques
...
+		TesResourcesBackendParameters: []string{
+			"",
```
adamstruck, line 114: Blank string?

Seems like this could be passed in via the config file.

tes/client.go
```diff
@@ -116,16 +118,15 @@ func (c *Client) CreateTask(ctx context.Context, task *Task) (*CreateTaskRespons
...
 	if err != nil {
 		return nil, fmt.Errorf("error marshaling task message: %v", err)
 	}
 
 	// Send request
 	u := c.address + "/v1/tasks"
-	hreq, _ := http.NewRequest("POST", u, &b)
-	hreq.WithContext(ctx)
+	hreq, _ := http.NewRequest("POST", u, bytes.NewReader(b))
+	// hreq.WithContext(ctx)
```
adamstruck, line 129:
```suggestion
	hreq = hreq.WithContext(ctx)
```

tests/core/basic_test.go
```diff
@@ -1148,17 +1145,23 @@ func TestConcurrentStateUpdate(t *testing.T) {
 			opts := &workerCmd.Options{TaskID: id}
 			w, err := workerCmd.NewWorker(ctx, c, log, opts)
 			if err != nil {
-				t.Fatal(err)
+				result = multierror.Append(result, err)
+				return
```
adamstruck, line 1149:
```suggestion
```

tests/funnel_utils.go
```diff
@@ -377,7 +378,9 @@ func (f *Funnel) StartServerInDocker(containerName, imageName string, extraArgs
 	cmd := exec.Command("docker", args...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
-	log.Info("Running command", "cmd", "docker "+strings.Join(args, " "))
+	log.Info("Running command!", "cmd", "docker "+strings.Join(args, " "))
+
+	// /test_tmp/funnel-test-1186572504/cj1diie9qp6c7392e3f0/tmp
```
adamstruck, line 383:
```suggestion
```

util/dockerutil/docker.go
```diff
@@ -34,7 +34,9 @@ func NewDockerClient() (*client.Client, error) {
 			}
 			// Error message example:
 			//   Error getting metadata for container: Error response from daemon: client is newer than server (client API version: 1.26, server API version: 1.24)
-			os.Setenv("DOCKER_API_VERSION", version[1])
+			// Hardcoding Docker version until version[1] returns valid version
+			// os.Setenv("DOCKER_API_VERSION", version[1])
+			os.Setenv("DOCKER_API_VERSION", "1.24")
```
adamstruck, line 39: This seems like something that will come back to bite you. 

You could try initializing the docker client the new way: https://pkg.go.dev/github.com/docker/docker/client#NewClientWithOpts

worker/docker.go
```diff
@@ -71,6 +71,7 @@ func (dcmd DockerCommand) Run(ctx context.Context) error {
 	args = append(args, dcmd.Command...)
 
 	// Roughly: `docker run --rm -i --read-only -w [workdir] -v [bindings] [imageName] [cmd]`
+	err = dcmd.Event.Info(fmt.Sprintf("args: %s", args))
```
adamstruck, line 74: Doesn't seem needed.

```suggestion
```

worker/worker.go
```diff
@@ -68,11 +68,14 @@ func (r *DefaultWorker) Run(pctx context.Context) (runerr error) {
 		event.Metadata(map[string]string{"hostname": name})
 	}
 
+	// TODO: Increment taskgroup waitgroup, defer done on waitgroup
+	// Decrement taskgroup waitgroup on exit to make sure that the final logging step is called
+	// before the worker is closed (and the backend database connection is closed).
+	//
```
adamstruck, line 74: I don't think this is still relevant?

database/mongodb/new.go
adamstruck: You could get rid of this if you use `--go-grpc_opt=paths=source_relative,require_unimplemented_servers=false`

cmd/task/get.go
```diff
@@ -19,23 +19,20 @@ func Get(server string, ids []string, taskView string, w io.Writer) error {
...
 	if err != nil {
 		return err
 	}
 
 	for _, taskID := range ids {
 		resp, err := cli.GetTask(context.Background(), &tes.GetTaskRequest{
 			Id:   taskID,
-			View: tes.TaskView(view),
+			View: taskView,
```
adamstruck, line 30: I’m surprised this accepts a string?

compute/scheduler/scheduler.go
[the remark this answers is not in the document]
lbeckman314: All good! I'm pretty new to the world of Go so have lots to learn in regards to it's error management and best practices. It's been nice learning from this code base!

webdash/src/Dashboard.js
adamstruck, line 1: The changes to how tags are expressed in v1.1 probably break the filtering capabilities in the web dashboard.

lbeckman314: That is 100% correct — working on adding support in the web dashboard tag filtering for `tag_key` and `tag_value` (over `tag` in v1.0).

adamstruck (commented):
(oops - I meant for this comment to be on the `tes.proto `file)

The `view` enum isn't used in the requests? That seems wrong. 

I know these changes were auto-generated from the openapi spec, but errors like this and the loss of all of the comments makes me wonder if that is the right approach.

tes/tes.proto
```diff
@@ -1,599 +1,241 @@
...
-
-  // OPTIONAL
-  //
-  // File content literal. 
-  // Implementations should support a minimum of 128 KiB in this field and may define its own maximum.
-  // UTF-8 encoded
-  //
-  // If content is not empty, "url" must be ignored.
-  string content = 6;
+enum view {
```
adamstruck, line 30: This enum isn't used in the requests which seems wrong.

I know these changes were auto-generated from the openapi spec, but errors like this and the loss of all of the comments makes me wonder if that is the right approach.

adamstruck: Maybe try using https://github.com/nytimes/openapi2proto?

adamstruck: Or another option: https://github.com/OpenAPITools/openapi-generator

adamstruck: The diagram you put together looks correct. 

> as it appears that the view enum is currently being used in a few places:
>   * Here is an example of the view enum value is being used in the Funnel CLI: [task.go]. (https://github.com/ohsu-comp-bio/funnel/blob/release/0.11.0-rc.1/cmd/task/task.go#L149)
>  * Here is where the client is creating the request to send to the sever with the same view enum value [client.go](https://github.com/ohsu-comp-bio/funnel/blob/release/0.11.0-rc.1/tes/client.go#L61)

The more important reason for using enums in the request definitions is so that the generated server will automatically reject requests with other values with an error like: 

```
$ curl localhost:8000/v1/tasks?view=FOOBAR

{
  "error": "FOOBAR is not a valid tes.TaskView",
  "code": 3,
  "message": "FOOBAR is not a valid tes.TaskView"
}
```

If the `view` is defined as a string (like it is in this PR), the validation is left up to you to implement. Here is what I get on your release branch:

```
$ curl localhost:8000/v1/tasks?view=FOOBAR
{
  "next_page_token":  "",
  "tasks":  [
    {
      "creation_time":  "",
      "description":  "",
      "executors":  [],
      "id":  "cl0tvebsaq988ichb7jg",
      "inputs":  [],
      "logs":  [],
      "name":  "",
      "outputs":  [],
      "resources":  null,
      "state":  "EXECUTOR_ERROR",
      "tags":  {},
      "volumes":  []
    }
  ]
}
```

(you can ignore the task error above, I didn't have docker running on my machine when I did this)

> Though this doesn't address the missing comments in tes.proto — I wonder if the alternatives you listed keep those in the generated output?

Yes, it looks like those tools generate comments in the proto file from the descriptions in the openapi definition.

adamstruck: Another note: I noticed that views now work a little differently. 

Before:

```
$ curl localhost:8000/v1/tasks/cl0u503saq98itrktup0?view=MINIMAL
{
  "id": "cl0u503saq98itrktup0",
  "state": "EXECUTOR_ERROR"
}
```

On 0.11.0-rc.1

```
$ curl localhost:8000/v1/tasks/cl0u503saq98itrktup0?view=MINIMAL
{
  "creation_time":  "",
  "description":  "",
  "executors":  [],
  "id":  "cl0u503saq98itrktup0",
  "inputs":  [],
  "logs":  [],
  "name":  "",
  "outputs":  [],
  "resources":  null,
  "state":  "EXECUTOR_ERROR",
  "tags":  {},
  "volumes":  []
}
```

Is this intentional? Why send the extra keys with empty values?

lbeckman314: Ah, definitely not intentional! Hmm...

# Resolution

Seems to be resolved by changing the `EmitUnpopulated` key to `false`  in `server.go`:

[server/server.go#L90](https://github.com/ohsu-comp-bio/funnel/blob/release/0.11.0-rc.1/server/server.go#L90)
```go
	m := protojson.MarshalOptions{
		Indent:          "  ",
		EmitUnpopulated: true,    <-- `true` results in keys being returned with empty values 
		UseProtoNames:   true,        (resulting in the client outputting extra keys)
	}
```

Will go ahead and update this value. 🫡

# Source of Change in gRPC Behavior

Looks like the behavior was changed in the  `grpc-gateway` package:
> We now emit default values for all fields
> https://grpc-ecosystem.github.io/grpc-gateway/docs/development/grpc-gateway_v2_migration_guide/#we-now-emit-default-values-for-all-fields

Related PR:
> runtime: make default marshaler emit default values
> https://github.com/grpc-ecosystem/grpc-gateway/pull/1377

Related issue:
> Support emitting default values in JSON
> https://github.com/grpc-ecosystem/grpc-gateway/issues/233

# Additional Resources

- [(runtime.JSONPb).UnmarshalOptions on pkg.go.dev](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson#MarshalOptions)

lbeckman314: >  If the view is defined as a string (like it is in this PR), the validation is left up to you to implement.

Ah, thank you for explaining that makes perfect sense — will go ahead and update the view from string to enum!

lbeckman314: Will also add an error handler in the server to reject requests with invalid `view` values (i.e. values not in the `view` enum):

> ```sh
> $ curl localhost:8000/v1/tasks?view=FOOBAR
> 
> {
>   "error": "FOOBAR is not a valid tes.TaskView",
>   "code": 3,
>   "message": "FOOBAR is not a valid tes.TaskView"
> }
> ```

adamstruck: > Will also add an error handler in the server

that isn't really needed if you make the changes to the generated proto.

lbeckman314: ## Potential Emit Defaults Issue

Thanks for your pointers on this @adamstruck! Getting things put together on this and running into a potential issue where we want to emit default/empty values for the BASIC and FULL views, but not for the MINIMAL view (where only 'id' and 'state' are emitted).

I'm not sure if protobuf supports this kind of logic using the same marshaler/unmarshaler [2]. Experimenting with a couple possible resolutions (listed below) but more than happy to implement any suitable changes on this.

## Possible Resolutions

- Custom Marshaler à la Grip ([marshaler.go](https://github.com/bmeg/grip/blob/main/server/marshaler.go))?

- Protobuf update with [openapi-generator](https://github.com/OpenAPITools/openapi-generator)?

- Add [FieldMask](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/FieldMask) to MINIMAL view?

- gRPC Interceptor?
  https://edgehog.blog/a-guide-to-grpc-and-interceptors-265c306d3773

- Revert to protobuf v2? 

- Potential future update to Go's JSON encoding?
  https://github.com/golang/go/discussions/63397#discussioncomment-7201222    

## Additional Resources

[1] Change from gRPC v1 to gRPC v2
https://grpc-ecosystem.github.io/grpc-gateway/docs/development/grpc-gateway_v2_migration_guide/

[2] Change from protobuf v2 to protobuf v3
https://github.com/protocolbuffers/protobuf/issues/1606

lbeckman314:
Cleaning things up and closing this PR in favor of the 0.11.0 release PR #749, all commits and changes here have now been included in the `release/0.11.0` branch.

Issue #748, opened 2023-12-05
Feature Request: Add support for native K8s Executor


lbeckman314:
Currently the Kubernetes support in Funnel is built around Docker-in-Docker (dind) as opposed to the native Kubernetes executor. This adds another layer of networking and resources that could otherwise be managed entirely by Kubernetes.

If possible migrate the existing dind Kubernetes task execution to the native k8s execution.

See [TESK's](https://github.com/elixir-cloud-aai/TESK) approach for task execution within Kubernetes as it already has built-in support for native task execution without the use of Docker-in-Docker.

PR #755, opened 2023-12-12
Adding support for native Kubernetes executor

KrKOo:
As stated in #748, running Docker-in-Docker for each job is not nearly as effective or secure as utilizing the Kubernetes API for the same purpose. This PR introduces optional native Kubernetes job execution. 

How does it work:
When the worker is configured to use the Kubernetes executor, it initializes the environment in the same way as when using Docker. Then instead of running a Docker container, it creates another Kubernetes Job.

TODO:
- [ ] Update the documention
- [ ] Implement continuous log collection from the job (if needed)


lbeckman314:
Thank you for the PR @KrKOo! We'll start reviewing it and hopefully get in approved for the release pipeline. 🚀

lbeckman314:
Apologies for the glacial pace with merging this issue — we just tested it locally and everything looks great, thank you for the contribution!

Rebasing onto latest development branch now...

PR #1373, opened 2026-04-01
Support Kubernetes retries

paulineribeyre:
## Description
- Support transitioning tasks out of a terminal state when they are being retried through Kubernetes restartPolicy / backoffLimit configurations.
- Update `ValidateTransition` logic so every case is clearly covered.
- When the executor fails, the worker catches the error and updates the task state, but the worker itself does not fail. This avoids unwanted worker retries when the issue is a failure in the user's script run by the executor.
- When a worker restarts, there was an error if the executor job already existed. Now the worker deletes and recreates the executor job.
- Update `waitForPodFinish` logic to give up if the executor pod has not appeared 5 min after the initial job creation, to log the pod state, and to log more details in case of error.

## Motivation and Context
We can now configure Funnel to retry in case of worker or executor error, for example:
```
Executor:
    restartPolicy: OnFailure
    backoffLimit: 1
Worker:
    restartPolicy: OnFailure
    backoffLimit: 2
```

## How Has This Been Tested?
Deployed and tested in K8s cluster

## Types of Changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist
- [x] I have tested that this feature locally.

### DIFF
--- a/tes/states.go
+++ b/tes/states.go
@@ -8,11 +8,11 @@ const (
 	Queued        = State_QUEUED
 	Initializing  = State_INITIALIZING
 	Running       = State_RUNNING
+	Paused        = State_PAUSED
 	Complete      = State_COMPLETE
 	ExecutorError = State_EXECUTOR_ERROR
 	SystemError   = State_SYSTEM_ERROR
 	Canceled      = State_CANCELED
-	Paused        = State_PAUSED
 )
 
 // TransitionError describes an invalid state transition.
@@ -67,7 +67,7 @@ func ValidateTransition(from, to State) error {
 
 	case ExecutorError, SystemError, Canceled, Complete:
 		// May not transition out of terminal state, except in the case of a retry.
-		// TODO configure if retries are allowed: maybe just check `Values.backoffLimit` > 0? nvm that only makes sense in k8s
+		// Whether to allow retries could be made into a configuration if needed
 		switch to {
 		case Queued, Initializing:
 			return nil
--- a/tes/states.go
+++ b/tes/states.go
@@ -50,28 +50,27 @@ func ValidateTransition(from, to State) error {
 		return nil
 
 	case Initializing:
-
-		switch to {
-		case Unknown, Queued:
+		if to == Unknown || to == Queued {
 			return &TransitionError{from, to}
 		}
 		return nil
 
 	case Running:
-
-		switch to {
-		case Complete, ExecutorError, SystemError, Canceled:
+		if to == Complete || to == ExecutorError || to == SystemError || to == Canceled {
 			return nil
 		}
 		return &TransitionError{from, to}
 
-	case ExecutorError, SystemError, Canceled, Complete:
-		// May not transition out of terminal state, except in the case of a retry.
+	case ExecutorError, SystemError:
+		// May not transition out of these terminal state, except in the case of a retry.
 		// Whether to allow retries could be made into a configuration if needed
-		switch to {
-		case Queued, Initializing:
+		if to == Queued || to == Initializing {
 			return nil
 		}
+		return &TransitionError{from, to}
+
+	case Complete, Canceled:
+		// May not transition out of these terminal states
 		return &TransitionError{from, to}
 
 	}
--- a/tes/states.go
+++ b/tes/states.go
@@ -6,13 +6,13 @@ import "fmt"
 const (
 	Unknown       = State_UNKNOWN
 	Queued        = State_QUEUED
+	Initializing  = State_INITIALIZING
 	Running       = State_RUNNING
-	Paused        = State_PAUSED
 	Complete      = State_COMPLETE
 	ExecutorError = State_EXECUTOR_ERROR
 	SystemError   = State_SYSTEM_ERROR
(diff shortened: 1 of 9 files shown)


tes/states.go
```diff
@@ -54,26 +54,28 @@ func ValidateTransition(from, to State) error {
 		switch to {
 		case Unknown, Queued:
 			return &TransitionError{from, to}
-		case Running, ExecutorError, SystemError, Canceled:
```
nss10, line 57: Maybe we can change the switch-case to if-else here?

tes/states.go
```diff
@@ -54,26 +54,28 @@ func ValidateTransition(from, to State) error {
...
-			return nil
 		}
+		return nil
 
 	case Running:
 
 		switch to {
-		case Unknown, Queued:
-			return &TransitionError{from, to}
```
nss10, line 65: Same here, can you replace switch-case with if-else

tes/states.go
```diff
@@ -54,26 +54,28 @@ func ValidateTransition(from, to State) error {
...
 			return nil
 		}
+		return &TransitionError{from, to}
 
 	case ExecutorError, SystemError, Canceled, Complete:
-		// May not transition out of terminal state.
+		// May not transition out of terminal state, except in the case of a retry.
+		// Whether to allow retries could be made into a configuration if needed
+		switch to {
+		case Queued, Initializing:
```
nss10: Are sure we are okay with tasks moving from completed/canceled to initializing?

tes/states.go
```diff
@@ -54,26 +54,28 @@ func ValidateTransition(from, to State) error {
...
+			return nil
+		}
 		return &TransitionError{from, to}
 
-	default:
```
nss10, line 74: I'm not sure if deleting the default case is advisable because this function kind of acts as a protection against state transitions that are not as per the design. So we allow only valid transitions; and through an error, for any transition that is not in the books.

worker/kubernetes.go
```diff
@@ -180,14 +175,30 @@ func (kcmd KubernetesCommand) Run(ctx context.Context) error {
 	var client = clientset.BatchV1().Jobs(kcmd.JobsNamespace)
 	_, err = client.Create(ctx, job, metav1.CreateOptions{})
 	if err != nil {
-		return &K8sSystemErr{
-			Reason:  "JobCreationFailed",
-			Message: "Failed to create Kubernetes job",
-			Err:     err,
+		// If the executor job already exists, delete and recreate it. This allows us to restart the
```
nss10, line 178: OK, I think this may be explains why you allowed transition from completed to initializing, because there could be a scenario where the executor completes, and just before it returns the flow back to the worker, the worker crashes. In order to handle worker restart, we are letting the executor job run again (at users' expense though). Right? 

I'm guessing you've already explored if there is any way where a worker can pick up where it left-off, based on the current task state?

worker/kubernetes.go
```diff
@@ -180,14 +175,30 @@ func (kcmd KubernetesCommand) Run(ctx context.Context) error {
...
+		// whole task in case of worker job error, even if the executor job is not configured to
+		// allow restarts.
+		if err.Error() == "jobs.batch \""+job.Name+"\" already exists" {
+			logger.Debug("Job already exists: recreating it", "jobName", job.Name)
+			deleteJob(ctx, clientset, client, job.Name, kcmd.JobsNamespace)
```
nss10, line 183: Can you make this 'executor jobs already exists', so that it's more explicit.

worker/kubernetes.go
```diff
@@ -322,37 +331,98 @@ func (kcmd KubernetesCommand) GetStderr() io.Writer {
...
+	// delete the job
+	var gracePeriod int64 = 0
+	var prop metav1.DeletionPropagation = metav1.DeletePropagationForeground
+	err := client.Delete(ctx, jobName, metav1.DeleteOptions{
+		GracePeriodSeconds: &gracePeriod,
+		PropagationPolicy:  &prop,
+	})
+	if err != nil {
+		return &K8sSystemErr{
+			Reason:  "JobCreationFailed",
```
nss10: Did you mean "job deletion failed"?

nss10 (changes requested):
Overall looks good. Just some minor corrections and questions.

tes/states.go
paulineribeyre: the default is duplicated, i removed one

worker/kubernetes.go
paulineribeyre: > if there is any way where a worker can pick up where it left-off, based on the current task state

I did not look into that, that sounds a lot more complex, it could be an improvement for later but it's not in scope right now

tes/states.go
paulineribeyre: I wasn't sure if it was possible for the state to skip "Queued" so i initially wrote "Queued, Initializing" and did not look into it again 🤔 @lbeckman314 might know

lbeckman314 (approved):

nss10 (approved):

lbeckman314:
- Merging + releasing in new tag only (no Helm Release as of yet)!

Issue #4, opened 2017-02-06
Change exported function/method/type to be private where possible


buchanae:
If it isn't necessary for the function/method/type to be exported, we should refactor it to be private.

Originally:
https://github.com/bmeg/task-execution-server/issues/11

PR #1396, opened 2026-04-22
fix: Skip ServiceAccount deletion if in use by other pods

lbeckman314:
# Overview 🌀

This PR updates the `DeleteServiceAccount` function to skip deleting ServiceAccounts if they're in use by any other pods.

This can result in downstream errors for workflows that expect ServcieAccounts to be user-scoped (rather than task-scoped).

## Current Behavior ❌

ServiceAccounts are deleted as soon as it's respective task completes.

## New Behavior ✔️

ServiceAccounts are now deleted as soon as it's respective task completes and it's not assigned to any other pods.

## Next Steps 🛠️

> [!IMPORTANT]
>
> - Verify that this new behavior is in-line with Gen3-Workflow's default behavior (e.g. via [`_WORKER_SA`](https://github.com/search?q=repo%3Auc-cdis%2Fgen3-workflow%20_WORKER_SA&type=code))
> - Add sequence diagram for the new behavior, including:
>   - Task submission
>   - ServiceAccount creation
>   - Task lifecycle
>   - ServiceAccount clean up

### DIFF
--- a/compute/kubernetes/resources/serviceaccount.go
+++ b/compute/kubernetes/resources/serviceaccount.go
@@ -68,6 +68,21 @@ func CreateServiceAccount(ctx context.Context, task *tes.Task, conf *config.Conf
 	return nil
 }
 
+func podsUsingServiceAccount(ctx context.Context, saName, namespace string, client kubernetes.Interface) ([]string, error) {
+	pods, err := client.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("listing pods using ServiceAccount %s: %v", saName, err)
+	}
+
+	var podNames []string
+	for _, pod := range pods.Items {
+		if pod.Spec.ServiceAccountName == saName {
+			podNames = append(podNames, pod.Name)
+		}
+	}
+	return podNames, nil
+}
+
 // DeleteServiceAccount deletes the ServiceAccount created for a task.
 func DeleteServiceAccount(ctx context.Context, taskID string, namespace string, client kubernetes.Interface, log *logger.Logger) error {
 	// ServiceAccount names are not available here without config, so we list by label.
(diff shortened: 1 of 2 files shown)


compute/kubernetes/resources/serviceaccount.go
nss10, line 90: I think we would miss an edge case here: 

1. Task t1 creates a service account with tags app=funnel;taskId=t1
2. Task t2 from the same user attempts to create sa for t2, but fails, since the above sa still exist, but t2 pods use this SA (since in gen3-workflow they are at user level) 
3. Task t1 completes, but since t2 is actively using this -- we no longer delete this SA
4. Task t2 completes, and it looks for SA with tags app=funnel;task=t2 -- _But it doesn't find any!_


@paulineribeyre -- considering this, I'm inclined to just cut the chase and implement task level service accounts for gen3-workflow as well? The motivation originally was to avoid duplicate service accounts being created with similar permissions and IRSA for each task from the same user, but maybe it is not worth going through all these special scenarios if these SAs are going to get created and deleted along with the tasks.

compute/kubernetes/resources/serviceaccount.go
```diff
@@ -68,6 +68,21 @@ func CreateServiceAccount(ctx context.Context, task *tes.Task, conf *config.Conf
 	return nil
 }
 
+func podsUsingServiceAccount(ctx context.Context, saName, namespace string, client kubernetes.Interface) ([]string, error) {
```
nss10, line 71: I know it would be useful to get a list of all the pods using a particular service account if we are building this as a feature, but if we are building only for this use case where we want to make sure we don't delete service accounts that are being used actively, we could rewrite this method to something like 
`isServiceAccountAttachedToPods` and look through all the pods for the service account, and if a match is found, break the loop and return True immediately.

paulineribeyre: +1 no need to go through the whole list

compute/kubernetes/resources/serviceaccount.go
```diff
@@ -68,6 +68,21 @@ func CreateServiceAccount(ctx context.Context, task *tes.Task, conf *config.Conf
...
+	pods, err := client.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{})
```
paulineribeyre, line 72: I think we might be able to pass listOptions instead of getting all the pods and filtering in the next block?
```
listOptions := metav1.ListOptions{
	FieldSelector: fmt.Sprintf("spec.serviceAccountName=%s", saName),
}
pods, err := clientset.CoreV1().Pods(namespace).List(ctx, listOptions)
```

compute/kubernetes/resources/serviceaccount.go
paulineribeyre: Can't we get the service account name from the task's `_WORKER_SA` tag instead of querying it through the task ID?

If not, from a quick google search i see it's possible to add a label to an existing service account, so we could add the task t2 label to fix #4.

PR #1407, opened 2026-05-07
Improve K8s resource cleanup and bugfixes related to Configmap creation

nss10:
## Description

This PR fixes three resource management issues in the Kubernetes backend related to 
1. ConfigMap templating
2. PersistentVolume scoping
3. ServiceAccount lifecycle.

---

## Motivation and Context

The `CreateConfigMap` function failed when a `ConfigMapTemplate` was configured because Go's `text/template` does not include an `indent` function — which the template relied on to correctly indent the worker config YAML under the `funnel-worker.yaml` block scalar key. Without proper indentation, the rendered output was invalid YAML and could not be decoded into a ConfigMap object.

- Registered `indent` as a custom template function
- Fixed template data key alignment (`Config` → `Data`, flat `Namespace` → `Release.Namespace`)

### Scope PersistentVolume deletion to the jobs namespace

PersistentVolumes are cluster-scoped resources, meaning a list or delete operation was not previously bounded to the current jobs namespace. Deletion now checks for a `namespace` label matching the configured `JobsNamespace` before proceeding, preventing accidental deletion of PVs belonging to other namespaces.

### Fix ServiceAccount deletion logic

ServiceAccounts in can be shared at the user level — multiple tasks from the same user reuse the same SA by passing a `_WORKER_SA` tag in the TES task definition. The previous cleanup logic was task-scoped and could either orphan SAs (when skipped because another task was active) or unsafely delete a shared SA still in use.

The new approach:
- Derives the SA name directly from the task, falling back to the conventional `funnel-worker-$namespace-$taskID` name if no `_WORKER_SA` tag is present
- For externally-managed SAs (tagged `_WORKER_SA`), checks for active pods before deletion and defers to the next reconciler pass if still in use
- Calls `DeleteServiceAccount` inline from `cleanResources`, eliminating the need to call `cleanOrphanedResources` on every reconciler tick and reducing unnecessary Kubernetes API calls

---

## Notes

Refactoring of the reconciler loop ([reconcile_modular](https://github.com/uc-cdis/funnel/tree/chore/reconcile_modular)) is tracked separately and will be submitted as a follow-up PR.

## Types of Changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [X] I have tested that this feature locally.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.


compute/kubernetes/resources/pv.go
```diff
@@ -83,6 +83,12 @@ func DeletePV(ctx context.Context, taskID string, client kubernetes.Interface, l
 			return fmt.Errorf("getting PV %s: %v", name, err)
 		}
 
+		if pv.Labels["namespace"] != namespace {
```
paulineribeyre, line 87:
```suggestion

		// PVs are not bound to any namespace. Funnel labels them with a namespace so that we do not delete PVs that were not generated by this instance of Funnel.
		if pv.Labels["namespace"] != namespace {
```

compute/kubernetes/resources/resources_test.go
```diff
@@ -381,7 +386,7 @@ func TestDeleteServiceAccount(t *testing.T) {
 		t.Fatalf("Failed to create test ServiceAccount: %v", err)
 	}
 
-	err = DeleteServiceAccount(context.Background(), testTaskID, namespace, fakeClient, l, false)
+	err = DeleteServiceAccount(context.Background(), testTaskID, namespace, fakeClient, l, nil)
```
paulineribeyre, line 389: `DeleteServiceAccount` doesn't use the `externalSA` param, should we just remove it?

compute/kubernetes/backend.go
```diff
@@ -422,6 +440,8 @@ func (b *Backend) reconcile(ctx context.Context, rate time.Duration, disableClea
 				}
 			}
 		}
+		b.CleanOrphanedResources(ctx)
```
paulineribeyre, line 443: This change seems to be doing the opposite of what's stated in the PR description?
> Calls DeleteServiceAccount inline from cleanResources, eliminating the need to call cleanOrphanedResources on every reconciler tick and reducing unnecessary Kubernetes API calls

compute/kubernetes/resources/serviceaccount.go
```diff
@@ -72,42 +73,66 @@ func CreateServiceAccount(ctx context.Context, task *tes.Task, conf *config.Conf
...
+}
+
+// DeleteServiceAccount deletes the ServiceAccount for a task.
+//
+// By default (no options), it targets the conventional SA name
+// funnel-worker-$namespace-$taskID and deletes it unconditionally.
+//
+// If opts.ExternalSA is true, the SA is treated as shared — deletion is skipped
+// if any active pod is still using it, and the reconciler will retry on the
+// next pass when the last pod finishes.
```
paulineribeyre: Should `ExternalSA` be renamed `SharedSA`? Or maybe even removed, and we can _always_ check whether the SA is in use before deleting it? This param seems like adding unnecessary complexity to me

compute/kubernetes/resources/pv.go
paulineribeyre, line 87: I don't see the label being added to the PV when it's created. Maybe it should be added [here](https://github.com/calypr/funnel/blob/4103475e492dec1fd853e9863d0caad15fc504c0/config/kubernetes/worker-pv.yaml#L5)?

nss10: Good point. It was added in the helm chart repo, but not here. Will make sure I include it, to stay consistent.

compute/kubernetes/resources/serviceaccount.go
nss10: One of the goals was to reduce number of calls made to K8s API. In cases where we know the SA is not external, I thought we could completely skip the check. Agreed that, this would not make any performance improvements for our use case, since we are always going to use external service account. 

I like the idea of renaming it to `SharedSA`, it is more descriptive.

compute/kubernetes/backend.go
nss10: Not really, `b.CleanOrphanedResources(ctx)` is ***not*** being called at every reconciler tick. Instead, we call it only once before we start the  reconciler loop. This is to cleanup any orphaned **_resources_** left behind in previous runs.

compute/kubernetes/resources/resources_test.go
nss10: Yeah, the last param `nil` corresponds to the [opts *DeleteServiceAccountOptions](https://github.com/uc-cdis/funnel/blob/e105eab09376775d4dc81a7ba3f96efb3078d924/compute/kubernetes/resources/serviceaccount.go#L93-L101) field.

compute/kubernetes/resources/pv.go
paulineribeyre: Oh gotcha. Maybe these duplicate files should just be removed from this repo (if they aren't used). I don't think they are up to date with the version in the helm-charts repo.
I was talking about this with @lbeckman314 the other day.

nss10: I did look into that. I believe they are needed to run unit tests.

compute/kubernetes/resources/resources_test.go
paulineribeyre: ooh 🤦‍♀️

lbeckman314 (approved):

lbeckman314 (approved):

paulineribeyre (approved):

lbeckman314:
- Unit Tests passing in fork: https://github.com/uc-cdis/funnel/actions/runs/25569726497

- Integration Tests blocked by current PR's not yet being merged

- Temp. enabling Bypass to merge this PR, then re-disabling Bypass for future PR's

PR #1410, opened 2026-05-11
Chore/reconcile upgrade

nss10:
## Description

* Modularize reconcile function such that it can be accessible to be invoked by a cron job, rather than just an internal ticker. 
* Add support to intelligently determine whether a job is supposed to be cleaned up when backoff Limit is set to a non zero value. 

## Types of Changes

## Checklist
- [ ] I have tested that this feature locally.


lbeckman314 (approved):
Approved in PR #1438 to allow Integration Tests to run!

PR #1413, opened 2026-05-12
fix: Improve K8s error diagnostics (`CreateContainerConfigError` + `BackoffLimit`)

lbeckman314:
# Overview 🌀

- **`CreateContainerConfigError` → `SYSTEM_ERROR`** *(worker-side)*: Executor pods stuck in a terminal waiting state (e.g. missing secret, invalid image name) are now detected in `waitForPodFinish` and immediately fail with `SYSTEM_ERROR` instead of hanging in `RUNNING` indefinitely. Pod Warning events are fetched and appended to the system log so users see the human-readable reason (e.g. `secret "foo" not found`) without needing `kubectl` access. Closes #75.

- **[Issue #73](https://ohsucomputationalbio.slack.com/lists/T04MS9FKW/F079AU1KCUX?record_id=Rec0B0QJ266M9) — BackoffLimit error message swallowing** *(reconciler-side)*: When a worker job exhausts its backoff retries, the reconciler now looks up the most recently failed pod and includes its exit code and termination reason as `executor_error` in the system log, alongside the existing `BackoffLimitExceeded` job conditions.

- **`K8sExecutorErr` reason fallback**: Kubernetes sets `Terminated.Reason` to `"Error"` (not empty) for plain non-zero exits. The error message now remaps both `""` and `"Error"` → `"ExitError"`, so logs show `(ExitError)` instead of the confusing `(Error): ''`.

# Test Steps ⚙️

- [x] Deploy with a task referencing a missing secret
  - **Expect:** Task transitions to `SYSTEM_ERROR`. System log contains `CreateContainerConfigError: executor pod has a terminal container waiting error: secret "..." not found` followed by `Failed: Error: secret "..." not found`

- [x] Deploy with `backoffLimit=0` and a failing command
  - **Expect:** System log `Exec error` shows `(ExitError)` instead of `(Error): ''`

- [x] Deploy with `backoffLimit>=1` and a failing command
  - **Expect:** Same as above — `Exec error` shows `(ExitError)`. The worker catches the first pod failure directly and reports `EXECUTOR_ERROR` before the job exhausts retries. The `Kubernetes job in FAILED state` + `executor_error` reconciler path is exercised when the **worker job itself** fails (e.g. OOMKilled worker pod), not the executor.

- [x] Deploy with a low `ramGb` and a memory-hungry command to OOMKill the executor pod
  - **Expect:** System log `Exec error` shows `exit code 137 (OOMKilled)` and task transitions to `EXECUTOR_ERROR`


compute/kubernetes/backend.go
```diff
@@ -384,6 +493,67 @@ func (b *Backend) isJobSchedulingTimedOut(ctx context.Context, jobName string, t
...
+		b.log.Error("reconcile: listing pods for failed job", "taskID", jobName, "error", err)
+		return ""
+	}
+
+	var latestFinish metav1.Time
+	var result string
+	for _, pod := range pods.Items {
+		for _, cs := range pod.Status.ContainerStatuses {
+			t := cs.State.Terminated
+			if t == nil {
```
nss10: Completely harmless, but can combine these two `if conditions`. 
```
        if t == nil || t.ExitCode == 0 {
            continue
        }
```

compute/kubernetes/backend.go
```diff
@@ -474,6 +646,7 @@ func (b *Backend) reconcile(ctx context.Context, rate time.Duration, disableClea
 			// List non-terminal tasks from Funnel's database
 			states := []tes.State{tes.State_QUEUED, tes.State_INITIALIZING, tes.State_RUNNING}
 			for _, s := range states {
+				fmt.Println("DEBUG: Reconciling tasks with state", s)
 				pageToken := ""
 				for {
 					lresp, err := b.database.ListTasks(ctx, &tes.ListTasksRequest{
```
nss10, line 650: Thanks to your debug log, I noticed a bug in the query. ListTasksRequest is returning all the tasks in the database, regardless of the task state 
```
DEBUG: Reconciling tasks with state QUEUED
DEBUG: Reconciling task d87vbq2pobks739q702g with state SYSTEM_ERROR
DEBUG: Reconciling task d87vb7sdj2sc73a3hjqg with state SYSTEM_ERROR
DEBUG: Reconciling task d87va0ipobks739q7020 with state SYSTEM_ERROR
...
...
DEBUG: Reconciling tasks with state INITIALIZING
DEBUG: Reconciling task d87vbq2pobks739q702g with state SYSTEM_ERROR
DEBUG: Reconciling task d87vb7sdj2sc73a3hjqg with state SYSTEM_ERROR
DEBUG: Reconciling task d87va0ipobks739q7020 with state SYSTEM_ERROR
...
```

nss10: Added a fix for this in reconcile_upgrade branch [commit](https://github.com/calypr/funnel/commit/7f60650ad8bab7d8b6a5b5785a0593aa9c17c3df). Feel free to include it in your branch.

compute/kubernetes/backend.go
```diff
@@ -384,6 +493,67 @@ func (b *Backend) isJobSchedulingTimedOut(ctx context.Context, jobName string, t
...
+		b.log.Error("reconcile: listing events for job", "taskID", jobName, "error", err)
+		b.log.Debug("assuming no FailedCreate events due to error listing events", "taskID", jobName)
+		return false, ""
+	}
+	if len(evList.Items) == 0 {
+		b.log.Debug("no FailedCreate events found for job", "taskID", jobName)
+		return false, ""
+	}
+	// Return the message from the most recent event.
+	latest := evList.Items[len(evList.Items)-1]
```
nss10: This logic is being very aggressive, not letting any new tasks to be created due to race conditions. 

When a new job is created, K8s also creates a service account for it, but it takes a while for sa to be created and accessible, but during this time, a FailedCreate event gets published in K8s, but the issue gets resolved immediately in under a couple of seconds. 

But here, since you are testing for at least one FailedCreate event, funnel is marking the task as `SYSTEM_ERROR` and failing the tasks. 

```
{"level":"debug","msg":"reconcile: worker job has FailedCreate event","ns":"kubernetes","reason":"Error creating: pods \"d87vv3i7qvgs739qbqh0-\" is forbidden: error looking up service account workflow-pods-qa-midrc-test/gen3wf-sai-dev-planx-pla-net-3-worker-sa: serviceaccount \"gen3wf-sai-dev-planx-pla-net-3-worker-sa\" not found","taskID":"d87vv3i7qvgs739qbqh0","time":"2026-05-22T07:08:15Z"}
```

```
$ kubectl get events -n workflow-pods-qa-midrc-test --sort-by='.lastTimestamp'  | grep d880bfapobks73eajr50
9s          Warning   FailedCreate       job/d880bfapobks73eajr50           Error creating: pods "d880bfapobks73eajr50-" is forbidden: error looking up service account workflow-pods-qa-midrc-test/gen3wf-sai-dev-planx-pla-net-3-worker-sa: serviceaccount "gen3wf-sai-dev-planx-pla-net-3-worker-sa" not found
8s          Normal    Scheduled          pod/d880bfapobks73eajr50-r4gwx     Successfully assigned workflow-pods-qa-midrc-test/d880bfapobks73eajr50-r4gwx to ip-172-20-12-144.ec2.internal
8s          Normal    SuccessfulCreate   job/d880bfapobks73eajr50           Created pod: d880bfapobks73eajr50-r4gwx
7s          Normal    Pulling            pod/d880bfapobks73eajr50-r4gwx     Pulling image "quay.io/ohsu-comp-bio/funnel:fix-system-error"
```

nss10 (changes requested):
`hasJobFailedCreateEvent` function did not let me create any new tasks couldn't test all the new features. Would be able to test further once that is fixed.

nss10:
> * **Ticket [#73](https://ohsucomputationalbio.slack.com/lists/T04MS9FKW/F079AU1KCUX?record_id=Rec0B0QJ266M9) — BackoffLimit error message swallowing**: When a job exhausts its backoff retries, the reconciler now also looks up the most recently failed pod and includes its exit code and termination reason as `executor_error` in the system log, alongside the existing `BackoffLimitExceeded` job conditions.

Incorrect hyperlink in the PR description

compute/kubernetes/backend.go
paulineribeyre: I wonder how the integration tests passed then 🤔

compute/kubernetes/backend.go
lbeckman314: Added in Commit [c089317](https://github.com/calypr/funnel/pull/1413/commits/c08931719ea7bb83fa8a6982c9936ce4319c7bb5)

compute/kubernetes/backend.go
lbeckman314: Updated in Commit [bddf1d9](https://github.com/calypr/funnel/pull/1413/commits/bddf1d94b7589d17eac65d24aa26de2eab10797d)

compute/kubernetes/backend.go
lbeckman314: Updated in Commit [37758f9](https://github.com/calypr/funnel/pull/1413/commits/37758f922860a56cafb5fdd2b29b84495926de63)

compute/kubernetes/backend.go
```diff
@@ -353,6 +353,115 @@ func (b *Backend) cleanResources(ctx context.Context, taskId string) error {
 	return errs
 }
 
+// terminalWaitingReasons lists container waiting reasons that will never self-
+// resolve and should be treated as a permanent failure.
+var terminalWaitingReasons = []string{
```
nss10: This could be inside the function `hasTerminalContainerWaitingError`, it is not used/needed outside of it.

compute/kubernetes/backend.go
```diff
@@ -353,6 +353,115 @@ func (b *Backend) cleanResources(ctx context.Context, taskId string) error {
...
+					msg := cs.State.Waiting.Message
+					if msg == "" {
+						msg = reason
+					}
+					return true, fmt.Sprintf("%s: %s", reason, msg)
+				}
+			}
+		}
+		// Also check init containers
+		for _, cs := range pod.Status.InitContainerStatuses {
```
nss10: Our worker or executor pods donot have any InitContainers in them. We don't need code for this. YAGNI principle :D

compute/kubernetes/backend.go
```diff
@@ -353,6 +353,115 @@ func (b *Backend) cleanResources(ctx context.Context, taskId string) error {
...
+
+// hasTerminalContainerWaitingError returns true if any pod belonging to the
+// given job has a container stuck in a waiting state whose reason is known to
+// be permanent (e.g. CreateContainerConfigError). These pods will never
+// transition to a running state on their own so the task must be failed early
+// rather than waiting for the Job's backoff limit to be exhausted.
+func (b *Backend) hasTerminalContainerWaitingError(ctx context.Context, jobName string) (bool, string) {
+	pods, err := b.client.CoreV1().Pods(b.conf.Kubernetes.JobsNamespace).List(ctx, metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("job-name=%s", jobName),
+	})
```
lbeckman314: - This is a costly call (possible to streamline by pulling in pods once and passing to methods as needed?)

compute/kubernetes/backend.go
```diff
@@ -504,8 +740,56 @@ func (b *Backend) reconcile(ctx context.Context, rate time.Duration, disableClea
...
+								continue
+							}
+
+							// Check for FailedCreate events on the Job itself. This catches
+							// cases where pod creation is rejected before a pod object is
+							// ever persisted (e.g. Pod Security Admission enforcement blocks
+							// the pod), so there are no pod container statuses to inspect.
+							// We require at least maxErrEventWrites events before treating
+							// the situation as permanent, to avoid false positives from
+							// transient API hiccups.
```
nss10: Can you explain what false positives do you mean here? I see `maxErrEventWrites` is an arbitrary hardcoded variable set to 2. A more robust check would be appreciated here.

compute/kubernetes/backend.go
```diff
@@ -504,8 +740,56 @@ func (b *Backend) reconcile(ctx context.Context, rate time.Duration, disableClea
 
 						jobName := j.Name
 						status := j.Status
+						fmt.Println("DEBUG: Job status for task", taskID, "is Active:", status.Active, "Succeeded:", status.Succeeded, "Failed:", status.Failed)
 						switch {
 						case status.Active > 0:
+							// Check for container waiting errors that will never self-resolve
+							// (e.g. CreateContainerConfigError). These keep the Job Active
+							// indefinitely, so we must detect and fail them explicitly.
+							if terminal, reason := b.hasTerminalContainerWaitingError(ctx, jobName); terminal {
```
nss10: When status.Active > 0, then maybe fetch pods using 
```
	pods, err := b.client.CoreV1().Pods(b.conf.Kubernetes.JobsNamespace).List(ctx, metav1.ListOptions{
		LabelSelector: fmt.Sprintf("job-name=%s", jobName),
	})
```
and then pass the pods list to `fetchPodWarningEvents` and `hasTerminalContainerWaitingError` and avoid repeating calling K8s apis to fetch the same information twice per each job.

compute/kubernetes/backend.go
```diff
@@ -353,6 +353,119 @@ func (b *Backend) cleanResources(ctx context.Context, taskId string) error {
...
+			for _, allowed := range podWarningEventReasons {
+				if ev.Reason == allowed {
+					key := ev.Reason + ":" + ev.Message
+					if _, dup := seen[key]; !dup {
+						seen[key] = struct{}{}
+						messages = append(messages, fmt.Sprintf("%s: %s", ev.Reason, ev.Message))
+					}
+					break
+				}
+			}
```
nss10: Can be replaced with slices.Contains no?

```suggestion
	if slices.Contains(podWarningEventReasons, ev.Reason) {
        key := ev.Reason + ":" + ev.Message
        if _, dup := seen[key]; !dup {
            seen[key] = struct{}{}
            messages = append(messages, fmt.Sprintf("%s: %s", ev.Reason, ev.Message))
        }
    }
```

compute/kubernetes/backend.go
```diff
@@ -353,6 +353,119 @@ func (b *Backend) cleanResources(ctx context.Context, taskId string) error {
...
+			reason := cs.State.Waiting.Reason
+			for _, terminal := range terminalWaitingReasons {
+				if reason == terminal {
+					msg := cs.State.Waiting.Message
+					if msg == "" {
+						msg = reason
+					}
+					return true, fmt.Sprintf("%s: %s", reason, msg)
+				}
+			}
```
nss10:
```suggestion
        if slices.Contains(terminalWaitingReasons, reason) {
                msg := cs.State.Waiting.Message
                if msg == "" {
                    msg = reason
                }
                return true, fmt.Sprintf("%s: %s", reason, msg)
        }
```

compute/kubernetes/backend.go
```diff
@@ -504,8 +740,56 @@ func (b *Backend) reconcile(ctx context.Context, rate time.Duration, disableClea
...
 						switch {
 						case status.Active > 0:
+							// Check for container waiting errors that will never self-resolve
+							// (e.g. CreateContainerConfigError). These keep the Job Active
+							// indefinitely, so we must detect and fail them explicitly.
+							if terminal, reason := b.hasTerminalContainerWaitingError(ctx, jobName); terminal {
+								b.log.Debug("reconcile: worker pod has terminal container waiting error", "taskID", jobName, "reason", reason)
+								b.event.WriteEvent(ctx, events.NewState(jobName, tes.SystemError))
+								errDetail := reason
+								if podEvents := b.fetchPodWarningEvents(ctx, jobName); podEvents != "" {
```
nss10: Can we replace this function call by calling function from kubernetes.go?
`fetchExecutorPodWarningEvents(ctx, b.client, b.conf.Kubernetes.JobsNamespace, jobName)`

worker/kubernetes.go
```diff
@@ -369,7 +383,69 @@ func (kcmd KubernetesCommand) GetStderr() io.Writer {
...
+	"ErrImagePull",     // explicit image-pull error
+	"ImagePullBackOff", // image pull back-off
+	"StartError",       // OCI runtime / entrypoint errors
+}
+
+// fetchExecutorPodWarningEvents returns deduplicated Warning events for pods
+// belonging to the executor job (label job-name=<jobName>), filtered to
+// reasons in podWarningEventReasons. Returns an empty string when nothing
+// useful is found.
+func fetchExecutorPodWarningEvents(ctx context.Context, clientset kubernetes.Interface, namespace, jobName string) string {
```
nss10: This function is not necessarily limited to Executor pods right? Can we rename it to `fetchPodWarningEvents` and use this same function even in reconcile logic? [here](https://github.com/uc-cdis/funnel/blob/73578c7965c0c2ea92d5f87c36a04ea9943e1206/compute/kubernetes/backend.go#L753C36-L753C36)

compute/kubernetes/backend.go
```diff
@@ -384,6 +497,126 @@ func (b *Backend) isJobSchedulingTimedOut(ctx context.Context, jobName string, t
...
+			if ts.IsZero() {
+				ts = metav1.Time{Time: ev.CreationTimestamp.Time}
+			}
+			if ts.After(latestFailed.Time) {
+				b.log.Debug("reconcile: FailedCreate resolved by later SuccessfulCreate", "taskID", jobName)
+				return 0, ""
+			}
+		}
+	}
+
```
nss10, line 534: Maybe add a timing check so that we only consider JobFailedCreate events if they happened after certain time threshold since the first failedCreate event? This would avoid race conditions when reconciler ran just after a job started and the JobFailedCreate occured 3 times, and SuccessfulCreate event did not get enough time to occur. Instead of relying on `maxErrEventWrites`, use a dedicated const variable `failedCreateThreshold` for this verification. 

This is something Claude gave me. 

```
// failedCreateThreshold is the minimum number of FailedCreate events required
// before the reconciler treats the job as persistently broken.
const failedCreateThreshold = 5

// minFailureSpan is the minimum duration between the first and last FailedCreate
// event required before the reconciler acts. This prevents false-positives from
// rapid-fire bursts in the first few seconds of a job's life.
const minFailureSpan = 20 * time.Second

// hasJobFailedCreateEvent returns the total FailedCreate count and the most
// recent failure message if the job has accumulated enough failures spread over
// enough real time, and no SuccessfulCreate has occurred after the last failure.
// Returns (0, "") if the failures look transient or self-resolved.
func (b *Backend) hasJobFailedCreateEvent(ctx context.Context, jobName string) (int, string) {
	ns := b.conf.Kubernetes.JobsNamespace

	allEvents, err := b.client.CoreV1().Events(ns).List(ctx, metav1.ListOptions{
		FieldSelector: fmt.Sprintf("involvedObject.name=%s", jobName),
	})
	if err != nil {
		b.log.Error("reconcile: listing events for job", "taskID", jobName, "error", err)
		return 0, ""
	}

	var (
		firstFailed  metav1.Time
		latestFailed metav1.Time
		latestMsg    string
		totalCount   int
		latestSuccess metav1.Time
	)

	for _, ev := range allEvents.Items {
		ts := eventTimestamp(ev)
		switch ev.Reason {
		case "FailedCreate":
			c := int(ev.Count)
			if c < 1 {
				c = 1 // brand-new singleton events have Count=0
			}
			totalCount += c
			if firstFailed.IsZero() || ts.Before(&firstFailed) {
				firstFailed = ts
			}
			if latestFailed.IsZero() || ts.After(latestFailed.Time) {
				latestFailed = ts
				latestMsg = ev.Message
			}
		case "SuccessfulCreate":
			if latestSuccess.IsZero() || ts.After(latestSuccess.Time) {
				latestSuccess = ts
			}
		}
	}

	if totalCount == 0 {
		return 0, ""
	}

	// If a SuccessfulCreate occurred after the last failure the Job controller
	// recovered on its own (e.g. a missing ServiceAccount was created moments
	// later by Helm); do not surface this as an error.
	if !latestSuccess.IsZero() && latestSuccess.After(latestFailed.Time) {
		b.log.Debug("reconcile: FailedCreate resolved by later SuccessfulCreate", "taskID", jobName)
		return 0, ""
	}

	// Require both a minimum count and a minimum span between first and last
	// failure before treating this as persistent. This avoids acting on:
	//   - a small number of failures that may still self-resolve
	//   - a rapid burst of failures all within the first few seconds
	failureSpan := latestFailed.Time.Sub(firstFailed.Time)
	if totalCount < failedCreateThreshold || failureSpan < minFailureSpan {
		b.log.Debug("reconcile: FailedCreate below persistence threshold, skipping",
			"taskID", jobName, "count", totalCount, "span", failureSpan)
		return 0, ""
	}

	b.log.Debug("reconcile: persistent unresolved FailedCreate events",
		"taskID", jobName, "count", totalCount, "span", failureSpan, "reason", latestMsg)
	return totalCount, latestMsg
}
```

compute/kubernetes/backend.go
lbeckman314: Updated in Commit [37758f9](https://github.com/calypr/funnel/pull/1413/commits/37758f922860a56cafb5fdd2b29b84495926de63)!

compute/kubernetes/backend.go
lbeckman314: Updated in Commit [9ad733ea](https://github.com/calypr/funnel/pull/1413/commits/9ad733ea)! (Debug prints removed; the ListTasks state filter fix was already included in a prior commit.)

compute/kubernetes/backend.go
lbeckman314: Yes! Updated in Commit [4c948330](https://github.com/calypr/funnel/pull/1413/commits/4c948330)

worker/kubernetes.go
lbeckman314: Renamed in Commit [4c94833](https://github.com/calypr/funnel/pull/1413/commits/4c94833)!

compute/kubernetes/backend.go
lbeckman314: Timing check would be a good feature, added in Commit [fadd9a6](https://github.com/calypr/funnel/pull/1413/commits/fadd9a6)

compute/kubernetes/backend.go
lbeckman314: Updated in Commit [4c94833](https://github.com/calypr/funnel/pull/1413/commits/4c948330) — `slices.Contains` is now used in `FetchPodWarningEvents`.

compute/kubernetes/backend.go
lbeckman314: Updated in Commit [fadd9a6](https://github.com/calypr/funnel/pull/1413/commits/fadd9a6c) — `maxErrEventWrites` replaced with `failedCreateThreshold` + `minFailureSpan` constants for a more robust and self-documenting check.

compute/kubernetes/backend.go
lbeckman314: Updated in Commit [2ea63b4](https://github.com/calypr/funnel/pull/1413/commits/2ea63b4a5bc0a61a6ea77c912aa89451b88fcf6c) — `terminalWaitingReasons` is now defined inside `hasTerminalContainerWaitingError` and no longer exposed at package scope.

compute/kubernetes/backend.go
lbeckman314: Updated in Commit [2ea63b4](https://github.com/calypr/funnel/pull/1413/commits/2ea63b4a5bc0a61a6ea77c912aa89451b88fcf6c) — InitContainers check removed.

compute/kubernetes/backend.go
lbeckman314: Updated in Commit [2ea63b4](https://github.com/calypr/funnel/pull/1413/commits/2ea63b4a5bc0a61a6ea77c912aa89451b88fcf6c) — pods are now fetched once in the reconciler and passed to both `hasTerminalContainerWaitingError` and `FetchPodWarningEvents`, eliminating the duplicate K8s API calls.

compute/kubernetes/backend.go
lbeckman314: Updated in Commit [2ea63b4](https://github.com/calypr/funnel/pull/1413/commits/2ea63b4a5bc0a61a6ea77c912aa89451b88fcf6c) — `FetchPodWarningEvents` now accepts a `*corev1.PodList` instead of fetching its own, and the reconciler passes the pre-fetched pod list to both functions.

compute/kubernetes/backend.go
lbeckman314: Updated in Commit [2ea63b4](https://github.com/calypr/funnel/pull/1413/commits/2ea63b4a5bc0a61a6ea77c912aa89451b88fcf6c) — `slices.Contains` is now used in `hasTerminalContainerWaitingError`.

compute/kubernetes/backend.go
```diff
@@ -626,6 +928,19 @@ func (b *Backend) CleanOrphanedResources(ctx context.Context) {
 
 	// TODO: Add Executor Jobs here beacause orphaned tasks can result in orphaned jobs
```
nss10: Remove the TODO:, also maybe close PR#1427, since the change is already included in this PR.

nss10 (approved):

lbeckman314:
Released in Helm Chart [v0.1.99-rc.38](https://github.com/calypr/helm-charts/pull/29/) [2026-05-10](https://github.com/calypr/funnel/commits/2026-06-10/)

PR #1357, opened 2026-03-19
Feat/k8s ovh improvements

geertvandeweyer:
## Description
This PR contains a couple of fixes made during deployment of Funnel/TES on OVHcloud.  They are mainly about mount-point consolidation (bypass the containerd 4096 byte limit for labels), add nerdctl to the docker image, and some stability tweaks. 

## Motivation and Context
Main motivation was the mountpoint consolidation required for containderd.  We use Cromwell upstream of TES/Funnel and each input file is provided as an individual mount.  In genomics pipelines, this rapidly blows up.   

## How Has This Been Tested?
- Tests were added *test.go files (and passed)
- Functional testing done on OVHcloud and AWS EKS. 

See for info on the deployment : https://geertvandeweyer.github.io/

## Types of Changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [ ] I have updated the documentation accordingly (link here).
- [x] I have tested that this feature locally.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [ ] Reviewer has tested this feature locally

PR #1421, opened 2026-05-20
fix: Resolve ServiceAccount deletion race condition (thanks @nss10!)

lbeckman314:
# Overview 🌀

This PR address [Issue #71 Clean Orphaned Resources as Go routine + PVC Finalizers](https://ohsucomputationalbio.slack.com/lists/T04MS9FKW/F079AU1KCUX?record_id=Rec0AUX3VPNNN) by adding improvements to how ServiceAccounts and related task resources are managed by Funnel.

## Current Behavior ⚠️

In Funnel [2026-04-21](https://github.com/calypr/funnel/commit/dc02e7bbfc35f3a582bc3a22c7cf7c3c39db592a), `cleanOrphanedResources` was moved to run at app startup. However, it was implemented as a blocking call, which causes health checks to fail when there are multiple tasks requiring cleanup. Additionally, after running for some time, the Kubernetes API becomes rate limited, causing the pod to crash.

**Proposed fixes:**

1. **Convert to a goroutine:** — Run `cleanOrphanedResources` asynchronously so it no longer blocks startup and health checks can proceed normally.

2. **Prevent Kubernetes API rate limiting** — Implement a batched deletion strategy with a sleep interval between batches to spread out API calls and stay within rate limits.

## New Behavior ✔️

- Added 5 second delay to avoid race conditions when cleaning up ServiceAccounts ([`serviceaccount.go`](https://github.com/calypr/funnel/pull/1421/changes#diff-46f318472e2f2caa19d4f03f588b89e4add69c346587c72c7d74d29a52e820dbR80-R81))

- Removed code cleaning up Namespace-scoped resources in favor of automatic [K8s Garbage Collection](https://kubernetes.io/docs/concepts/architecture/garbage-collection/) via [Ownership fields](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/) ([`backend.go`](https://github.com/calypr/funnel/pull/1421/changes#diff-6b8bdaf0017913b075ada768862a9fcce4ec277e19e8fb5692e2dc2e01347e0aL299-L346))

## Breaking Changes? ❌

> [!CAUTION]
>
> Yes! This update currently breaks the [TTL config](https://github.com/calypr/funnel/compare/develop...uc-cdis:funnel:chore/reconcile_upgrade#diff-fef184bddb3fc509046cac20ae5fc42907a2906c7aaf6832656a170126853bf2L117-R125) first introduced by [PR #1357](https://github.com/calypr/funnel/pull/1357).
>
> This feature is being re-introduced as part of PR #1410 to avoid breaking existing configurations.


compute/kubernetes/backend.go
```diff
@@ -296,29 +296,6 @@ func (b *Backend) cleanResources(ctx context.Context, taskId string) error {
 		b.log.Error("deleting Job", "error", err)
 	}
 
-	// Delete PVC
-	err = resources.DeletePVC(ctx, taskId, b.conf.Kubernetes.JobsNamespace, b.client, b.log)
```
nss10, line 300: @lbeckman314 — Deferring the deletion of these resources to Kubernetes using ownerReferences sounds good. However, we should handle that transition in a separate PR.

It also just occurred to me: what happens to cleanOrphanedResources? If there are any pending resources left behind by older server runs or due to previous deletion errors, our logic might track them but never actually clean them up.

Kubernetes garbage collection will continuously retry deleting these resources until they are gone, so we don't need to worry about transient API errors or rate limiting. 

Given this, we should do one of two things:
1. Make these changes in a separate PR, and mark this change as a breaking change.
2. Update the `cleanOrphanedResources` method so it no longer looks for these specific resources as "orphaned," since Kubernetes will handle the cleanup natively.

compute/kubernetes/backend_test.go
```diff
@@ -174,11 +188,9 @@ spec:
 		t.Error("expected Job to be deleted, but it still exists")
 	}
 
-	// Verify that the ConfigMap was deleted
-	_, err = fakeClient.CoreV1().ConfigMaps(conf.Kubernetes.JobsNamespace).Get(context.Background(), configMapName, metav1.GetOptions{})
-	if err == nil {
-		t.Error("expected ConfigMap to be deleted, but it still exists")
-	}
+	// ConfigMap deletion is handled by Kubernetes garbage collection via ownerReferences,
```
nss10: Instead of Configmap, look for a resource which is actually being cleaned up by `cleanResources` -- say PV. But as I mentioned above, it has to be in a PR of its own.

nss10 (changes requested):

compute/kubernetes/backend.go
lbeckman314: Updated in PR #1423!

compute/kubernetes/backend_test.go
lbeckman314: Updated in PR #1423!

compute/kubernetes/backend.go
nss10: Hi @lbeckman314 , I was suggesting #1423 to be a PR pointed to `develop`, such that it includes all changes related to resource cleanup. Keep #1421 only about handling Service Account deletion race condition. All logic regarding resource cleanup should be handled in #1423 only.

nss10 (approved):

PR #1439, opened 2026-06-24
feat: Add initial `funnel k8s cleanup` command

lbeckman314:
## Description

Adds an initial `funnel kubernetes cleanup` command that deletes orphaned Funnel-managed Kubernetes resources (PVs, ServiceAccounts, executor jobs) whose task is no longer present or is in a terminal state. It is designed to run inside the cluster as a CronJob — the "external reconciler" — so that resource cleanup is decoupled from the Funnel Server lifecycle and multiple server replicas do not race to clean the same resources.

The command reuses the running server's configuration and connects to the same cluster; it does **not** stand up a second compute backend or require a worker job template.

### Changes

- **`cmd/kubernetes/kubernetes.go`**
  - New `funnel kubernetes cleanup` subcommand. Opens only the configured Funnel database (no HTTP/gRPC server) and runs `CleanOrphanedResources`.
  - Add a `k8s` alias for the `kubernetes` command group, matching the backend name which already accepts both `kubernetes` and `k8s`.
- **`compute/kubernetes/backend.go`**
  - New `NewCleanupBackend` constructor: a cleanup-only backend that, unlike `NewBackend`, does not require a worker job template and does not start the reconcile goroutine. It only needs a k8s client, a DB reader, and config.
  - Resolves the target namespace from config, falling back to `JobsNamespace` and then the pod's ServiceAccount namespace.
- **`util/k8sutil/k8s.go`**
  - Add `InClusterNamespace`, which reads the namespace from the pod's mounted ServiceAccount, so the CronJob works without an explicitly configured namespace.

## Motivation and Context

K8s resource cleanup currently runs inside the Funnel Server via the reconcile goroutine, coupling it to the server lifecycle and causing replicas to race. This command is the CLI entry point for running that cleanup as an independent, schedulable CronJob.

Related PRs:

- calypr/funnel#1438 and calypr/funnel#1410 — modularize the reconciler so a single pass (`reconcileOnce`) can be invoked externally. Once those land, this command should call the full single-pass reconcile rather than only `CleanOrphanedResources` (tracked by a `TODO` in `cmd/kubernetes/kubernetes.go`).
- calypr/helm-charts#22 — adds the CronJob that invokes `funnel kubernetes cleanup --config /etc/config/funnel-server.yaml` and derives its schedule from `Kubernetes.ReconcileRate`.

## How Has This Been Tested?

- Builds and `go vet` pass locally.
- `funnel k8s --help` / `funnel kubernetes --help` resolve to the same command group (alias verified).
- The command is intended to run **in-cluster** (it uses in-cluster config). It can be exercised by triggering the CronJob from calypr/helm-charts#22, e.g. `kubectl create job --from=cronjob/funnel-cleanup-<namespace> funnel-cleanup-manual`.

## Types of Changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have tested that this feature locally.

### DIFF
--- a/cmd/kubernetes/kubernetes.go
+++ b/cmd/kubernetes/kubernetes.go
@@ -27,8 +27,9 @@ import (
 
 // Cmd is the root "funnel kubernetes" command.
 var Cmd = &cobra.Command{
-	Use:   "kubernetes",
-	Short: "Funnel Kubernetes management commands.",
+	Use:     "kubernetes",
+	Aliases: []string{"k8s"},
+	Short:   "Funnel Kubernetes management commands.",
 }
 
 func init() {
@@ -67,11 +68,13 @@ is decoupled from the server lifecycle and multiple replicas do not race.`,
 				return fmt.Errorf("opening database: %v", err)
 			}
 
-			// TODO: Why are we building a new backend here?
-			// Build the K8s backend (connects to the cluster via in-cluster config).
-			// We pass a no-op event writer since this command only deletes resources
-			// and never needs to emit task state events.
-			backend, err := k8sbackend.NewBackend(ctx, conf, reader, &events.Logger{Log: log}, log)
+			// Build a cleanup-only K8s backend. This connects to the same cluster as
+			// the running Funnel Server (via in-cluster config) but, unlike the full
+			// server backend, does not require a worker job template and does not start
+			// the reconcile goroutine — cleanup only reads the database and deletes
+			// orphaned resources. The event writer just logs, since cleanup never emits
+			// task state events.
+			backend, err := k8sbackend.NewCleanupBackend(conf, reader, &events.Logger{Log: log}, log)
 			if err != nil {
 				return fmt.Errorf("initializing kubernetes backend: %v", err)
 			}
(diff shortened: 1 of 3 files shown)


compute/kubernetes/backend.go
```diff
@@ -840,6 +875,18 @@ func (b *Backend) isResourceCleanupNeeded(ctx context.Context, taskID string) (b
...
 
+// ReconcileOnce performs a single reconciliation pass over Funnel-managed Kubernetes
+// resources. It is intended to be invoked by an external scheduler (e.g. a Kubernetes
+// CronJob configured via the Helm chart's ReconcileRate value) so that reconciliation
+// is decoupled from the Funnel server lifecycle and multiple server replicas do not
+// race to reconcile the same resources.
+//
+// This is currently a stub that simply logs; the full reconciliation logic is added in
+// https://github.com/calypr/funnel/pull/1438.
+func (b *Backend) ReconcileOnce() {
```
matthewpeterkort: So this function needs to go away and then upgrade the existing function to be exported

cmd/kubernetes/kubernetes.go
```diff
@@ -67,17 +68,27 @@ is decoupled from the server lifecycle and multiple replicas do not race.`,
...
 
 			log.Info("Starting orphaned resource cleanup",
 				"namespace", conf.Kubernetes.JobsNamespace)
+
+			// Delete Funnel-managed resources whose task is gone or terminal.
+			//
+			// TODO: Once the modular reconciler lands (calypr/funnel#1438, #1410)
+			// this should call the full single-pass reconcile (reconcileOnce) so the
+			// CronJob also reconciles task/job state, not just orphaned resources.
 			backend.CleanOrphanedResources(ctx)
```
matthewpeterkort: I think it is alsotime to update this

matthewpeterkort (changes requested):

cmd/kubernetes/kubernetes.go
nss10: Yeah! We just need to call `ReconcileOnce` here.

compute/kubernetes/backend.go
nss10: Correct. And ensure when the `k8s cleanup` command is called -- we need to trigger a reconciler run.

PR #1150, opened 2025-02-14
`feature/plugins` → `develop`: Plugin Support (WIP)

lbeckman314:
# ⚙️ Overview 

> Adapted from [*RPC-based plugins in Go*](https://eli.thegreenplace.net/2023/rpc-based-plugins-in-go) by [Eli Bendersky](https://eli.thegreenplace.net/) and [go-plugin examples](https://github.com/hashicorp/go-plugin/tree/main/examples/basic) 🚀

This PR adds initial support for Plugins using the [`go-plugin`](https://github.com/hashicorp/go-plugin) package from [HashiCorp](https://github.com/hashicorp).

In this setup, the Plugin handles all user authentication, with the Server having no "knowledge" or record of user credentials/tokens (e.g. [`example-users.csv`](./authorizer/example-users.csv)).

## 🔄 Sequence Diagram

> Created with https://sequencediagram.org ([_source_](https://github.com/ohsu-comp-bio/funnel/blob/feature/plugins/plugins/sequence-diagram.txt))

proposed-auth-design

# ⚡ Quick Start 

## 1. Start the Server 

```console
➜ cd plugins

➜ make
Building ./server...OK

➜ ./server
Listening on http://localhost:8080
```

## 2. Send Requests

In another terminal, send the request using of the examples below —

<details>
 <summary>Required Inputs ⚙️</summary>
 
| Input | Sent in...           | Example                                  |
|-------|----------------------|------------------------------------------|
| User  | Authorization Header | `Authorization: Bearer Alyssa P. Hacker` |
| Task  | Request Body         | `@example-tasks/hello-world.json`        |

</details>

<details>
 <summary>Example Users ⚙️</summary>
 
All ["Authorized" users](https://en.wikipedia.org/wiki/Structure_and_Interpretation_of_Computer_Programs#Characters) may be found in the "User Database" — [`example-users.csv`](https://github.com/ohsu-comp-bio/funnel/blob/feature/plugins/plugins/example-users.csv):
> - Alyssa P. Hacker, a Lisp hacker
> - Ben Bitdiddle
> - Cy D. Fect, a "reformed C programmer"
> - Eva Lu Ator
> - Lem E. Tweakit
> - Louis Reasoner, a loose reasoner

Here the use of a simple text file to contain users is to show how a plugin can be written and integrated into Funnel — real world use cases might involve more complex mechanisms for tracking users (e.g OAuth)...

</details>

### Example: Authorized User

[...]

PR #1447, opened 2026-08-26
Show credentials inside Credential where they need to be

### LINKED ISSUE #1446: Credentials in RPCClient need to be inside a `Credential` field now, but the documentation
In commit 4215dcf from #1150, the `User` and `Password` for `RPCClient` were moved inside of a `Credential` field. See:
https://github.com/calypr/funnel/commit/4215dcf0a23b369222390ebb1853060fd64f57ec#diff-aa49e8740efa4f7fbc76c17f3b2a68301a17f7c05f90bb2e4da90a3075a2c467L96-R97

The Github UI is baffled about the tag order, but this commit is in the current v0.11.14 release. The oldest right-looking version it is in is v0.11.8.

`User` and `Password` live happily inside `Credentials` to this day:
https://github.com/calypr/funnel/blob/97d1df559ba31b0c17a0ab1f6794e826c36c4450/config/config.proto#L91

But nobody has yet updated the documentation, which still shows them at the old location as direct members of `RPCClient`:
https://github.com/calypr/funnel/blob/97d1df559ba31b0c17a0ab1f6794e826c36c4450/website/content/docs/security/basic.md?plain=1#L41-L45

When you have them there as shown in the documentation, you get an exciting error message on server startup, like:
```
Error: error processing config: failed to parse config at path /builds/adamnovak/toil_batch_system_tes/funnel.conf: errs: failed to unmarshal JSON with protojson: proto: (line 1:589): unknown field "User"
```

(This message is in terms of line and character numbers that, notably, aren't the same as the lines and characters in the config file itself.)

The documentation should be changed to match the code. It might also help to get the line numbers working for errors in the config file, and to note in the documentation that old configs have to be updated for this change as of whatever version it was first released in.

adamnovak:
## Description
This updates the example `RPCClient` configuration section to have the `User` and `Password` fields in the right place.

## Motivation and Context
This should fix #1446. In #1150 the `User` and `Password` fields of `RPCClient` were moved inside a field named `Credential`, but the documentation's example configuration still shows them at their old location.

## How Has This Been Tested?
I made my configuration look like this and started Funnel v0.11.14 and the server came up instead of rejecting the configuration.

## Types of Changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [X] I have updated the documentation accordingly (link here).
- [X] I have tested that this feature locally.

I don't think the documentation is under test, right? Otherwise this would have been caught already.

I wrote this in the Github web editor, and I'm not set up with a real dev environment to run the test suite.

### DIFF
--- a/website/content/docs/security/basic.md
+++ b/website/content/docs/security/basic.md
@@ -40,8 +40,9 @@ so you will also need to configure the RPC client.
 
 ```yaml
 RPCClient:
-  User: funnel
-  Password: abc123
+  Credential:
+    User: funnel
+    Password: abc123
 ```
 
 Make sure to properly protect the configuration file so that it's not readable


matthewpeterkort:
Thanks for spotting this. I have pulled in your commit from this branch into https://github.com/calypr/funnel/pull/1450 which aims to address your other issues as well.

Issue #1449, opened 2026-08-26
Funnel doesn't give me a usable temporary directory unless I ask nicely


adamnovak:
**Describe the bug**

I'm trying to help bring Toil's TES plugin (last working with the Funnel 0.10 series) up to date with current Funnel v0.11.14). See: https://github.com/adamnovak/toil_batch_system_tes/pull/6

I've gotten to the point where my containers will run, but my Python code inside the container imports `dill`, and Dill tries to get a temporary directory from the Python builtin `tempfile` module on import, and this fails because it can't find a usable temporary directory.

```
WARNING  toil_batch_system_tes.tes_batch_system:tes_batch_system.py:452 Log from failed executor: Traceback (most recent call last):
  File "/usr/local/bin/_toil_contained_executor", line 3, in <module>
    from toil.batchSystems.contained_executor import executor
  File "/usr/local/lib/python3.10/dist-packages/toil/batchSystems/contained_executor.py", line 27, in <module>
    from toil.batchSystems.abstractBatchSystem import EXIT_STATUS_UNAVAILABLE_VALUE
  File "/usr/local/lib/python3.10/dist-packages/toil/batchSystems/abstractBatchSystem.py", line 30, in <module>
    from toil.deferred import DeferredFunctionManager
  File "/usr/local/lib/python3.10/dist-packages/toil/deferred.py", line 23, in <module>
    import dill
  File "/usr/local/lib/python3.10/dist-packages/dill/__init__.py", line 33, in <module>
    from .session import (
  File "/usr/local/lib/python3.10/dist-packages/dill/session.py", line 25, in <module>
    TEMPDIR = pathlib.PurePath(tempfile.gettempdir())
  File "/usr/lib/python3.10/tempfile.py", line 454, in gettempdir
    return _os.fsdecode(_gettempdir())
  File "/usr/lib/python3.10/tempfile.py", line 447, in _gettempdir
    tempdir = _get_default_tempdir()
  File "/usr/lib/python3.10/tempfile.py", line 362, in _get_default_tempdir
    raise FileNotFoundError(_errno.ENOENT,
FileNotFoundError: [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/']
```

When I add `/tmp` as a volume in my TES task, this works, and `dill` can be imported inside the container. But I don't want the temporary files from my (only) executor to be on storage that must be preserved between executors; I want them to be on scratch storage for just that executor.

[...]

Issue #1448, opened 2026-08-26
Funnel 0.11.14 can't detect resources on macOS


adamnovak:
**Describe the bug**

I downloaded `funnel-darwin-arm64-v0.11.14.tar.gz` and extracted the `funnel` binary from it.

When I do `./funnel node run`, my log starts with:

```
FLAG CONF BEFORE MERGE CALL:
node                 error detecting resources
error                Error detecting cpu cores: no such file or directory
nodeID               01a0400e-1b16-7236-a85d-3d4825f2273d
src                  node.go:github.com/ohsu-comp-bio/funnel/compute/scheduler.NewNodeProcess:34
time                 2026-08-26T17:51:01-04:00

node                 Starting node
nodeID               01a0400e-1b16-7236-a85d-3d4825f2273d
time                 2026-08-26T17:51:01-04:00
```

If I actually configure Funnel and start the server, the node can connect to the server, but it still complains about being unable to detect CPU cores. Even if I pass `--Node.Resources.Cpus 2` on the command line, I still get this error about not being able to detect CPU cores.

The node doesn't seem interested in running any tasks even with the CPU core count set manually, I think because the resource-detecting function stops trying to detect resources after the first thing it can't detect, and so if it can't detect CPU cores it doesn't go on to try and detect memory.

https://github.com/calypr/funnel/blob/97d1df559ba31b0c17a0ab1f6794e826c36c4450/compute/scheduler/util.go#L33-L36

**To Reproduce**
1. Get Funnel v0.11.14 for macOS.
2. Start the node with any sort of `funnel node run`
3. Observe the log on standard error

**Expected behavior**

Funnel should be able to detect CPU cores on macOS.

When it can't, Funnel should continue and try to detect other kinds of resource.

**Screenshots**
N/A

**Environment (please complete the following information):**
 - Environment: `Darwin GI-ANOVAK1-01 25.5.0 Darwin Kernel Version 25.5.0: Tue Jun  9 22:26:46 PDT 2026; root:xnu-12377.121.10~1/RELEASE_ARM64_T8103 arm64 arm Darwin`
 - Version: `v0.11.14`

**Additional context**
CPU detection seems to work OK on Linux.

adamnovak:
Since this machine is specifically an M1 Mac, this is probably https://github.com/shirou/gopsutil/issues/1598 AKA https://github.com/shirou/gopsutil/issues/1000, which is supposedly fixed by using `v3` or later of the `gopsutil` library.

[...]

PR #1450, opened 2026-08-27
fix: improve executor portability and configuration defaults

matthewpeterkort:
## Summary

- restore writable `/tmp` in the default read-only Docker container with an executor-local tmpfs
- suppress that tmpfs when a TES task explicitly mounts `/tmp` or an ancestor
- update host resource detection to `gopsutil/v4` and detect CPU, memory, and disk independently
- correct RPC credential nesting in the configuration docs and examples, with parsing regression coverage

Adam Novak's three-line documentation change from #1447 is preserved as its own authored commit. That PR remains open for the contributor/maintainer to close separately.

## Verification

- `go test ./... -count=1` (413 tests across 83 packages)
- Apple Silicon host resource probe through the scheduler suite
- Docker smoke test: read-only root with writable `/tmp` tmpfs
- template tests for implicit tmpfs and explicit `/tmp` volume suppression

Refs #1446
Refs #1448
Refs #1449


matthewpeterkort:
@codex review

matthewpeterkort:
@adamnovak this PR should address your current issues.

